PR #1, opened 2020-08-26
Story/tp65870 home page

bakerac4:


pep/app/pods/components/page/sidebar/widgets/template.hbs
```diff
@@ -0,0 +1,15 @@
+<div class="p-3 d-flex">
+    <a href="" {{on "click" this.toggleAll}}>
+        {{t (if this.showCloseAll 'relatedInfo.batchClose' 'relatedInfo.batchOpen')}}
+    </a>
```
billdami, line 4: This should be a <button> to be more semantically correct (and removes the need for the `event.preventDefault()` in the action), you can keep the link-like appearance by doing 
```hbs
<Button @type="link" class="p-0" .../>
```
Couple of other more design-related things i noticed here;
- The toggle button looks odd just hanging out at the top, i think now that we have a parent action control there, we should add back the "Related Info" h5 heading I had previously removed, and align the toggle to the right of the heading, to anchor it a bit better.
- The cards going right to the edges of the sidebar makes them look less "card-like", I think at least a little padding around everything (`p-2` seems like a decent amount) cleans that up a bit
- I am still pretty strongly against the action-icon-in-card-header pattern here 😛   I adjusted the html in devtools to show my suggested pattern in the screenshot below to visualize it better. I think thats the better way to go in this case for a bunch of reasons, but willing to bring it to a broader discussion if you want.

Current layout (for reference):
current

Adjusted:
adjusted

pep/app/pods/components/page/sidebar/widgets/component.ts
```diff
@@ -0,0 +1,37 @@
...
+export interface PageSidebarWidgetArgs {
+    toggleIsOpen: (widget: WIDGET) => void;
+    openWidgets: WIDGET[];
+}
+
+export default class PageSidebarWidgets extends Component<PageSidebarWidgetsArgs> {
+    @tracked openWidgets: WIDGET[] = [WIDGET.WHATS_NEW, WIDGET.MOST_CITED, WIDGET.MOST_VIEWED];
+
+    get showCloseAll() {
+        return this.openWidgets.length > 2;
```
billdami, line 16: from a user perspective this felt a bit weird/unexpected for that toggle to change on `> 2` open. I'm not sure which way the logic should go but i think it feels more natural if its on `> 0` open or closed, e.g. either show "Show all" if at least one widget is closed, otherwise show "Close all". Or, show "Close all" if at least one widget is open, otherwise show "Show all".

pep/app/pods/components/page/sidebar/widgets/component.ts
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    @action
```
billdami, line 19: missing doc blocks on methods

pep/app/pods/components/power-select/journal/component.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    /**
+     * Loads the initial page of options
+     * @return {Array}
+     */
+    @action
+    async loadInitialPage() {
+        if (!this.fastboot.isFastBoot) {
```
billdami, line 60: use the `@dontRunInFastboot` decorator instead of this if block, and then i think you can remove the service injection itself too.

pep/translations/en-us.json
```diff
@@ -279,12 +282,36 @@
         }
     },
     "mostCited": {
-        "title": "Most cited journal articles",
-        "period": "Period"
+        "pageTitle": "Most cited journal articles",
+        "period": "Period",
+        "periodSelect": "Seelct a period",
```
billdami, line 287: typo

pep/translations/en-us.json
```diff
@@ -243,6 +243,9 @@
         }
     },
     "relatedInfo": {
+        "title": "Related Info",
+        "batchClose": "Close All",
+        "batchOpen": "Show All",
```
billdami, line 248: these labels should probably use sentence casing to follow the formatting of action/button labels elsewhere in the app

pep/tests/integration/pods/components/page/sidebar/widgets/component-test.ts
```diff
@@ -0,0 +1,26 @@
...
+        assert.equal(this.element.textContent?.trim(), '');
+
+        // Template block usage:
+        await render(hbs`
+      {{#page/sidebar/widgets}}
+        template block text
+      {{/page/sidebar/widgets}}
+    `);
+
+        assert.equal(this.element.textContent?.trim(), 'template block text');
```
billdami, line 24: looks like a few of these tests are failing

pep/app/styles/components/_ember-table.scss
```diff
@@ -0,0 +1,23 @@
...
+            border-radius: 0 !important;
+        }
+    }
+    th:last-child,
+    td:last-child,
+    th:first-child,
+    td:first-child {
+        padding-left: 0;
+    }
+}
```
billdami, line 23: i think you might not need a bunch of these styles, or any, now that the addon's sass styles should be imported. but if any of the color based styles are still needed, they should use existing variables, or create new ones which need to go in both _variables-theme.scss, and themes/tomato.scss for theming purposes.

pep/app/pods/most-cited/controller.ts
```diff
@@ -7,30 +7,48 @@ import { inject as service } from '@ember/service';
...
+import Journal from 'pep/pods/journal/model';
+import { PERIODS, PossiblePeriodValues } from 'pep/constants/sidebar';
 
 export default class MostCited extends Controller {
     @service loadingBar!: LoadingBarService;
     @service fastbootMedia!: FastbootMediaService;
     @service sidebar!: SidebarService;
+    @service intl!: IntlService;
 
     queryParams = ['author', 'title', 'sourcename', 'period'];
```
billdami, line 20: i think we should look at this later in another PR, but i think its not the best UX that as the user types in filter fields, the URL query params update immediately. There is the browser history pollution issue (tho that can be partly fixed with `replace: true` configs), but also, i think its slightly nicer to not update the query params until a new search has been run w/the new filter values (basically the behavior of the current Search page)

pep/app/pods/components/tables/most-cited/component.ts
```diff
@@ -8,45 +13,112 @@ interface TablesMostCitedArgs {
...
+            ? [
+                  {
+                      id: '0',
+                      valuePath: 'title',
+                      name: this.intl.t('mostCited.table.title'),
+                      isFixedLeft: false,
+                      width: 200,
+                      staticWidth: 200,
+                      cellComponent: 'tables/cell/most-viewed-publication',
+                      isSortable: true
```
billdami, line 38: are we hitting the endpoint for server-side sorting when a column is sorted? If not i dont think we should enable client-side sorting since these tables are paginated right? Otherwise as the user scrolls down to load more records, all subsequent pages would either not be sorted, or there would be some weird re-arranging of rows as they scrolled down, to put them in the right order, I think.

pep/app/pods/most-cited/controller.ts
```diff
@@ -7,30 +7,48 @@ import { inject as service } from '@ember/service';
...
+     * Eventually we should use journal IDS to search on instead of the title so that way we can
+     * reload them if coming from a pasted link
+     *
+     * @memberof MostCited
+     */
+    get sourcename() {
+        return this.journal?.title ?? '';
+    }
+    set sourcename(value: string) {
+        console.log(value);
```
billdami, line 39: i would just comment out the console.log if this isnt fully done yet

pep/app/pods/most-cited/nav/template.hbs
```diff
@@ -1,113 +1,68 @@
...
-                                }}
-                            </p>
-                            <p class="mb-0">
-                                {{t "search.form.tooManyResults.instructions"}}
-                            </p>
-                        </Alert>
+                            <PowerSelect::Journal
+                                @selected={{this.journal}}
+                                @onChange={{this.updateJournal}}
+                            />
```
billdami, line 47: we need to look at cleaning up the styling of the power selects to match the styles of the other inputs/native selects, but we can do that later.

pep/app/pods/components/page/sidebar/widgets/component.ts
bakerac4: I think it would make the most sense at greater than or equal to 2. If you have one opened, you can just as easily close it the normal way. if you have two opened, we should show close all.

pep/app/styles/components/_ember-table.scss
bakerac4: We need these due to the sub columns

pep/app/pods/most-cited/controller.ts
bakerac4: Agreed completely

pep/app/styles/_common.scss
```diff
@@ -1,6 +1,8 @@
 // ---------------------------
 // Vendor / Template
 // ---------------------------
+@import 'variables';
+@import 'variables-theme';
```
billdami, line 5: theses cant be imported here, as its used across all themes, and both are already imported in `styles/app.scss` and `styles/themes/tomato.scss`.

pep/app/styles/_variables-theme.scss
```diff
@@ -30,3 +31,4 @@ $headings-font-family: Muli, -apple-system, BlinkMacSystemFont, 'Segoe UI', Robo
 $hr-border-color: rgba(#000, 0.06);
 $dropdown-border-color: #dbdbdb;
 $dropdown-border-radius: 0.21rem;
+$table-border: #e9e9e9;
```
billdami, line 34: shouldn't need to redefine `$white` as it already exists, and `$table-border` would also need to be defined in `styles/themes/tomato.scss` too, otherwise the sass build will fail. Although I think we should probably remove this variable altogether, and just use the existing bootstrap `$table-border-color` variable where needed in _ember-table.scss.

pep/app/styles/components/_ember-table.scss
```diff
@@ -2,10 +2,10 @@
 .ember-table {
     thead {
         th {
-            border-top: 1px solid #fff;
+            border-top: 1px solid $white;
         }
         tr:nth-child(2) > th {
-            border-top: 1px solid #e9e9e9;
+            border-top: 1px solid $table-border;
```
billdami, line 8: as mentioned above, i would just use the existing `$table-border-color` here to match the rest of the table's border color exactly.

pep/app/pods/components/page/sidebar/widgets/whats-new/template.hbs
```diff
@@ -2,11 +2,16 @@
...
+    class="mb-2 page-widget"
     {{did-insert this.onElementInsert}}
     ...attributes
 as |Panel|>
-    <Panel.header @title={{t "relatedInfo.widgets.whatsNew.title"}} />
+    <Panel.header>
+        {{t "relatedInfo.widgets.whatsNew.title"}}
+        <span class="widget-icon" {{on "click" this.openModal}}>
+            <FaIcon @icon="envelope" @prefix="fal"/>
+        </span>
```
billdami, line 13: scared

pep/app/pods/components/page/sidebar/widgets/template.hbs
```diff
@@ -1,15 +1,18 @@
-<div class="p-3 d-flex">
-    <a href="" {{on "click" this.toggleAll}}>
+<div class="p-3 d-flex justify-content-between align-items-center">
+    <h5 class="m-0">{{t "relatedInfo.title"}}</h5>
+    <Button @type="link" class="p-0" {{on "click" this.toggleAll}}>
         {{t (if this.showCloseAll 'relatedInfo.batchClose' 'relatedInfo.batchOpen')}}
-    </a>
+    </Button>
 </div>
```
billdami, line 6: probably should have `px-2 py-3` instead of `p-3` for this container to match the widgets' horizontal padding. 

use `@action={{this.toggleAll}}` instead of `{{on "click" this.toggleAll}}` on `<Button>`'s.

pep/app/pods/components/page/sidebar/widgets/component.ts
billdami: yeah i think changing this to `>= 2` is fine too.

pep/app/pods/most-cited/controller.ts
bakerac4: Leaving this in to allow for the green builds you mentioned

pep/app/pods/most-cited/nav/template.hbs
bakerac4: Thats fine - though I don't necessarily agree that they should look exactly like native selects because they are so different.

pep/types/ember-concurrency/index.d.ts
```diff
@@ -0,0 +1,117 @@
+declare module 'ember-concurrency' {
```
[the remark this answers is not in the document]
bakerac4: Short answer is - no I don't think that's possible. I switched to that and it still has a peer dep on ember-concurrency-decorators so these are still being shown.

pep/app/pods/components/tables/most-cited/component.ts
bakerac4: We are hitting the server when we filter

pep/app/pods/most-cited-loading/template.hbs
```diff
@@ -0,0 +1,3 @@
+<Page::Content>
+    <Loading::Read::Document />
```
billdami, line 2: Note for future work, but we should make a `<Loading::Table />` component to use for these data table pages. We have a custom skeleton placeholder component for tables that i made extending from the addon, and have it in use in CC/J&R/etc

pep/app/pods/components/page/sidebar/widgets/whats-new/template.hbs
billdami: my list of concerns with this pattern is still unresolved 😛, but in the interest in making progress and getting this stuff merged in, we can punt on that. as i have time tho, i do intend to push a branch w/my proposed pattern of either my original design, or at least rework this new pattern a bit to alleviate some of my concerns.

pep/app/styles/_variables-theme.scss
billdami: looks like `$table-border` is no longer being used anywhere

pep/app/pods/components/page/sidebar/widgets/whats-new/template.hbs
bakerac4: I added the items you wanted per your screenshot. I kept the icons as well to make it easier for a user to do the action related to the widgets when they were closed. We can discuss whenever you want

pep/package.json
```diff
@@ -82,6 +83,9 @@
         "ember-cli-typescript-blueprints": "^3.0.0",
         "ember-cli-uglify": "^3.0.0",
         "ember-composable-helpers": "^4.2.1",
+        "ember-concurrency": "^1.2.1",
+        "ember-concurrency-decorators": "^2.0.1",
```
billdami, line 87: do we still need ember-concurrency-decorators now that you have ember-concurrency-ts?

bakerac4: Due to not being able to get rid of ember-concurrency-decorators as its a peer dep of ember-concurrency-ts, I didn't change anything to use ember-concurrency-ts. But Im leaving both in so we can play around with things

PR #2, opened 2020-08-26
Release/002

billdami:


pep/app/utils/csv.ts
```diff
@@ -0,0 +1,11 @@
...
+import { urlForDocumentQuery } from 'pep/pods/document/adapter';
+import { serializeQueryParams } from 'pep/utils/url';
+import DS from 'ember-data';
+
+export const csvUrl = (store: DS.Store, queryParams: QueryParamsObj) => {
+    delete queryParams.limit;
+    queryParams.download = true;
+    const url = urlForDocumentQuery(store, { queryType: queryParams.queryType });
+    return `${url}?${serializeQueryParams(queryParams)}`;
+};
```
billdami, line 11: @bakerac4  i probably would throw this in the existing `utils/url.ts` util file, since its probably not likely we'll have too many or any more utils relating to csvs in this project.

Interesting use of the document query adapter to avoid duplicating all that logic. I'd live with it, but I dont _love_ that we have to pass in the `store` context to use this, makes the util a little less portable/standalone. I probably would've just extracted all the url logic from the document adapter's `urlForQuery` method itself, made that a generic util w/o needing the `Store` directly, and then use it both the original adapter, and here. But we can do that in a follow up PR if you want.

pep/app/pods/most-viewed/controller.ts
```diff
@@ -11,6 +11,10 @@ import IntlService from 'ember-intl/services/intl';
 import Journal from 'pep/pods/journal/model';
 import { PERIODS, PossiblePeriodValues } from 'pep/constants/sidebar';
 import { QueryParams } from 'pep/hooks/useQueryParams';
+import { buildQueryParams } from '@gavant/ember-pagination/utils/query-params';
+import { urlForDocumentQuery } from 'pep/pods/document/adapter';
```
billdami, line 15: @bakerac4  doesnt look like u are using `urlForDocumentQuery` in this file. (you should be getting yellow squigglys and warnings in the build for unused imports I think?)

pep/app/pods/most-cited/controller.ts
```diff
@@ -11,6 +11,10 @@ import IntlService from 'ember-intl/services/intl';
 import Journal from 'pep/pods/journal/model';
 import { PERIODS, PossiblePeriodValues } from 'pep/constants/sidebar';
 import { QueryParams } from 'pep/hooks/useQueryParams';
+import { buildQueryParams } from '@gavant/ember-pagination/utils/query-params';
+import { urlForDocumentQuery } from 'pep/pods/document/adapter';
```
billdami, line 15: @bakerac4  doesnt look like u are using `urlForDocumentQuery` in this file either

pep/app/pods/most-cited/controller.ts
```diff
@@ -106,6 +110,17 @@ export default class MostCited extends Controller {
     updateJournal(journal: Journal) {
         this.searchQueryParams.journal = journal;
     }
+
+    @action
```
billdami, line 114: @bakerac4  missing doc block

pep/app/pods/most-viewed/controller.ts
```diff
@@ -100,6 +105,17 @@ export default class MostViewed extends Controller {
     updateJournal(journal: Journal) {
         this.searchQueryParams.journal = journal;
     }
+
+    @action
```
billdami, line 109: @bakerac4 missing doc block

PR #7, opened 2020-08-27
Story/tp65872 right panel widget actions

billdami:
@bakerac4 here's my proposed tweaks to the panel header actions to address some of the concerns i mentioned before: 

- Refactored the panel header markup to to use semantic `<button>`s for both the main toggle and the optional secondary action, make them siblings instead of layered on top of each other, and increase click target size for secondary action, to improve usability on mobile.
- Give secondary action's icon the same text color as links, to better indicate they are an action
- Add `aria-label` and tooltips to secondary actions
- Replace the "table" icons with "file-search" (searching for "view all icon" in google gave this icon and the "eye" icon as the top matches, so seemed a bit more indicative as to its action).
- Added the "Update subscriptions" text-based action link inside the what's new widget to match the other components (I think we should always still have this in-content "call to action" control)

Resulting in this for the UI:
Screen Shot 2020-08-27 at 1 22 00 PM


billdami:
> Looks slightly bulky to me, but overall I think its a good compromise.

Really? When looking at them in-person with the rest of the app they actually look more subtle to me, the blue link color helps with that i think, and at the same time its easier to tell they are a separate action rather than just decorative. But yea we can definitely get more feedback as we go along and circle back to the sidebar stuff as needed.

anyway, looks like you approved this but didn't merge?

PR #8, opened 2020-08-27
Add initial about dialog

bakerac4:


pep/app/pods/components/modal-dialogs/user/about/component.ts
```diff
@@ -0,0 +1,27 @@
...
+    @service theme!: ThemeService;
+
+    /**
+     * Update the current theme
+     * @param {String} newThemeId
+     */
+    @action
+    updateTheme(newThemeId: string) {
+        this.theme.updateTheme(newThemeId);
+    }
```
billdami, line 26: looks like all this theme related stuff should be ripped out.

pep/app/pods/components/modal-dialogs/user/about/template.hbs
```diff
@@ -0,0 +1,24 @@
...
+    <Modal.body>
+        <div class="mb-2">
+            <div>{{t "about.client"}}</div>
+            <div class="text-muted small">{{@options.clientBuildVersion}}</div>
+        </div>
+        <div class="mb-2">
+            <div>{{t "about.server"}}</div>
+            <div class="text-muted small">{{@options.serverInformation.text_server_version}}</div>
+            <div class="text-muted small">{{@options.serverInformation.dataSource}}</div>
+        </div>
```
billdami, line 12: in my stickler effort to strive for more semantic markup in this project 😛 (since accessibility is a big element), content like this should probably be structured more like:
```hbs
<h5 class="mb-0">{{t "about.server"}}</h5>
<p class="text-muted small">
    {{@options.serverInformation.text_server_version}}<br>
    {{@options.serverInformation.dataSource}}
</p>
```

pep/app/pods/components/modal-dialogs/user/about/template.hbs
```diff
@@ -0,0 +1,24 @@
...
+    </Modal.body>
+    <Modal.footer>
+        <div class="d-flex justify-content-between w-100 align-items-center">
```
billdami, line 15: you should be able to remove the need for this inner div container, and just do this right on the footer component:

```hbs
<Modal.footer class="justify-content-between align-items-center">
    ...
</Modal.footer>
```
and should no longer need `d-flex` or `w-100` in this case too.

pep/app/pods/components/page/nav/component.ts
```diff
@@ -37,6 +49,25 @@ export default class PageNav extends Component<PageNavArgs> {
...
     }
 
+    /**
+     * Open the about modal dialog
+     *
+     * @returns Promise<Void>
+     * @memberof PageDrawer
+     */
+    @action
+    async openAboutModal() {
```
billdami, line 59: why are you `await`ing the promise in this `openAboutModal`, but not in the page/drawer version? I think we should probably be consistent either way, and if we await it, show the `loadingBar` while its waiting for the response.

I know its not a ton of code right now, but don't love that this logic is duplicated in 2 places, you could just make an `about` service (see my `auth` service for example), which would be a nicer place for that `ServerStatus` interface to live too probably rather than in this component.

pep/app/pods/components/page/nav/component.ts
```diff
@@ -37,6 +49,25 @@ export default class PageNav extends Component<PageNavArgs> {
...
+        try {
+            const result = await this.ajax.request<ServerStatus>('Session/Status');
+            return this.modal.open('user/about', {
+                serverInformation: result,
+                clientBuildVersion: ENV.buildVersion
+            });
+        } catch (error) {
+            return error;
```
billdami, line 67: i think you'd want to `throw` the error here instead of resolving the process successfully with an error. Or in this case you could remove the `try/catch` altogether, since this is the default behavior of async functions (unless you add `loadingBar` in which case you'd call `loadingBar.hide()` in a `finally` block.

pep/tests/integration/pods/components/modal-dialogs/user/about/component-test.ts
```diff
@@ -0,0 +1,26 @@
...
+        assert.equal(this.element.textContent?.trim(), '');
+
+        // Template block usage:
+        await render(hbs`
+      {{#modal-dialogs/user/about}}
+        template block text
+      {{/modal-dialogs/user/about}}
+    `);
+
+        assert.equal(this.element.textContent?.trim(), 'template block text');
```
billdami, line 24: this test is prob failing since the modal doesnt have a block form. I'm trying to see if we can always have completely green builds 100% for any code pushed to a release/ or master branches on this project.

I know not having tests turned on CI or blocking merges until they pass makes this harder to enforce, but I think its doable.

pep/app/pods/components/page/nav/component.ts
billdami: ideally, you could also use `this.notifications.error(...)` to display an error toast in the catch, in case we couldnt retrieve the server info for w/e reason.

pep/tests/integration/pods/components/modal-dialogs/user/about/component-test.ts
bakerac4: Nothing failing except the random issues with GenericTaskProperty that we have talked about before

pep/app/pods/components/page/nav/component.ts
bakerac4: I do like the idea of having a nice home for that interface, but I feel like creating a service for this is overkill. You'll probably end up with a large abundance of services that only do minimal work and will make it harder to keep a mental model of the project. Id lean towards keeping it like this unless is grows?

bakerac4: Perhaps we should have a service that groups all of this stuff together. Like a `MenuService` or something. That way your not creating one off services all the time for such small functionality

billdami: yeah i totally agree that having separate services for everything is not the best either, that was my one reservation in suggesting that.

Actually, heres a better solution and probably what we should be doing here; since both the page/nav and page/drawer components are children of the application controller, we can move the `openAboutModal` action into the controller, and pass it down as an arg to both components, `<Page::Nav @openAboutModal={{this.openAboutModal}} />`. 

And the application/controller is a bit of a nicer more central place for the `ServerStatus` interface to live too I think.

billdami: Yeah, having a more generic service for more than just "about" was gonna be my other suggestion before i thought of the controller/arg solution. But id be fine with either.

pep/app/pods/application/controller.ts
```diff
@@ -7,10 +7,27 @@ import SessionService from 'ember-simple-auth/services/session';
 
 import LoadingBarService from 'pep/services/loading-bar';
 import { SEARCH_TYPE_EVERYWHERE, SearchTermValue } from 'pep/constants/search';
+import AjaxService from 'pep/services/ajax';
+import NotificationService from 'ember-cli-notifications/services/notifications';
+
+import Modal from '@gavant/ember-modals/services/modal';
+import ENV from 'pep/config/environment';
+export interface ServerStatus {
```
billdami, line 15: not necessary now, but maybe the best thing would be to have like a `types/api/..` file for these kinds of interfaces, especially if we end up having other raw ajax request endpoints we hit later that need their own interfaces too.

pep/app/pods/application/controller.ts
```diff
@@ -96,6 +113,29 @@ export default class Application extends Controller {
...
+    async openAboutModal() {
+        try {
+            this.loadingBar.show();
+            const status = await this.ajax.request<ServerStatus>('Session/Status');
+            return this.modal.open('user/about', {
+                serverInformation: status,
+                clientBuildVersion: ENV.buildVersion
+            });
+        } catch (error) {
+            this.notifications.error(error);
```
billdami, line 133: probably doubtful the api is returning us a string we can use directly as an error message here, so I'd probably pass in `this.intl.t('serverErrors.unknown.unexpected')` to the `error()` call.

pep/app/pods/components/page/drawer/component.ts
```diff
@@ -7,13 +7,16 @@ import ModalService from '@gavant/ember-modals/services/modal';
...
+import AjaxService from 'pep/services/ajax';
+interface PageDrawerArgs {
+    openAboutModal: () => Promise<void>;
+}
 
 export default class PageDrawer extends Component<PageDrawerArgs> {
     @service drawer!: DrawerService;
     @service session!: SessionService;
     @service auth!: AuthService;
+    @service ajax!: AjaxService;
```
billdami, line 19: probably no longer need this injection

pep/app/pods/components/page/nav/component.ts
```diff
@@ -7,14 +7,21 @@ import ModalService from '@gavant/ember-modals/services/modal';
...
+}
 
 export default class PageNav extends Component<PageNavArgs> {
     @service modal!: ModalService;
     @service session!: SessionService;
+    @service ajax!: AjaxService;
     @service auth!: AuthService;
     @service drawer!: DrawerService;
+    @service notifications!: NotificationService;
+    @service loadingBar!: LoadingBarService;
```
billdami, line 24: could get rid of these injections now too

PR #9, opened 2020-08-28
Story/tp65735 auth

bakerac4:


pep/app/authenticators/pep.ts
```diff
@@ -21,10 +21,10 @@ export default class PepAuthenticator extends BaseAuthenticator {
     async authenticate(username: string, password: string) {
         const params = serializeQueryParams({
             grant_type: 'password',
-            username,
-            password
+            UserName: username,
+            Password: password
         });
-        const result = await this.ajax.request(`Session/Login?${params}`);
+        const result = await this.ajax.request(`https://padstest.zedra.net/PEPSecure/api/v1/Authenticate?${params}`);
```
billdami, line 27: you prob just didnt get to this yet, but we'll want to make this base url for the auth endpoints a new env var, like `AUTH_BASE_URL` and then import it here from `ENV`. to add new env vars you need to update a couple files:

- add it to every `.env-*` file in the root (leaving the value blank for staging/prod)
- pep/config/dotenv.js
- pep/app/config/environment.d.ts
- pep/config/environment.js

.env-production
```diff
@@ -6,6 +6,7 @@ API_NAMESPACE=v2
 API_DATA_NAMESPACE=Database
 API_ADMIN_NAMESPACE=Admin
 API_ERRORS_ENDPOINT=/error-events
+AUTH_BASE_URL=
```
billdami, line 9: @lhorne-gavant just making sure you're aware of this new env var we'll want to add to our build scripts

pep/app/services/ajax.ts
```diff
@@ -56,7 +56,6 @@ export default class AjaxService extends Service {
      */
     async request<T>(url: string, options: RequestInit = {}): Promise<T> {
         setProperties(options, {
-            credentials: 'include', //NOTE: only needed if cookies must be sent in API requests (which are needed for auth right now)
```
billdami, line 59: just noting that i think we will ultimately need this left in (as alan mentioned the apis will want to be sending/sharing cookies for some other stuff), so if if its causing issue with PaDs, we may need to have him address that. For the time being, i'd probably just comment this out instead of deleting it.

pep/app/authenticators/pep.ts
```diff
@@ -21,10 +25,12 @@ export default class PepAuthenticator extends BaseAuthenticator {
...
-            password
+            UserName: username,
+            Password: password
+        });
+        const result = await this.ajax.request(`${ENV.authBaseUrl}/Authenticate?${params}`, {
+            headers: this.authenticationHeaders
         });
-        const result = await this.ajax.request(`Session/Login?${params}`);
         return result;
     }
```
billdami, line 35: not sure if @adistasio talked to you about this yet, but apparently we are never doing OAuth now, so we'll need some to add some logic to the authenticator that automatically invalidates the local session when the returned session expires.

I'll most likely look somewhat similar to the `_scheduleAccessTokenRefresh` logic we have in our oauth-based authenticators (see CC or JRE), but instead of refreshing, itll just call `invalidate()` once the expire date is hit. This timer would most likely need to be kicked off both in `authenticate()` and `restore()`.

If we want to do this in a follow up PR, thats probably fine too.

pep/app/pods/application/adapter.ts
```diff
@@ -23,6 +23,16 @@ export interface ApiServerErrorResponse {
...
 
+export interface PepSecureAuthenticatedData {
+    HasSubscription: boolean;
+    IsValidLogon: boolean;
+    IsValidUserName: boolean;
+    ReasonId: number;
+    ReasonStr: string;
+    SessionId: string;
+    authenticator: string;
+}
```
billdami, line 34: this might be better off living in `types/pep/api.d.ts`

pep/app/pods/application/adapter.ts
```diff
@@ -66,10 +76,12 @@ export default class Application extends DS.RESTAdapter.extend(DataAdapterMixin,
      * @returns Object
      */
     get headers() {
-        const headers = { client_id: ENV.clientId } as any;
+        const headers = { 'client-id': ENV.clientId } as any;
         if (this.session.isAuthenticated) {
-            const { access_token } = this.session.data!.authenticated;
-            headers['Authorization'] = `Bearer ${access_token}`;
+            // We are converting to unknown because session data is specified as something
+            // completely different by the addon
```
billdami, line 82: we've custom defined this `authenticated ` session data interface ourselves, in `types/ember-simple-auth/services/session.d.ts`. So its probably better to just change it there directly, to `PepSecureAuthenticatedData` so we don't need to convert here or anywhere else.

pep/app/services/ajax.ts
```diff
@@ -19,14 +20,16 @@ export default class AjaxService extends Service {
...
         const headers = {} as any;
         // api auth token is sent in cookies
-        // if (this.session.isAuthenticated) {
-        //     const { access_token } = this.session.data!.authenticated;
-        //     headers['Authorization'] = `Bearer ${access_token}`;
-        // }
+        if (this.session.isAuthenticated) {
+            // We are converting to unknown because session data is specified as something
+            // completely different by the addon
+            const { SessionId } = (this.session.data!.authenticated as unknown) as PepSecureAuthenticatedData;
```
billdami, line 30: same here as the last comment, if we just adjust our already custom session data interface, we won't need to convert it everywhere.

billdami (changes requested):

billdami:
@bakerac4 does this fix #27?

pep/app/pods/application/adapter.ts
bakerac4: I'd tend to disagree with this on the basis that eventually we won't be specifying those types as they will provide them and in that case we will have to do this anyways

billdami: @bakerac4 totally agree on that point (that we can remove all ember-simple-auth specific types) once they have types defined themselves. 

But in this case, our `authenticated` interface is custom and API-dependent (the existing `SessionAuthenticatedData` interface we defined is for OAuth2, so wouldn't even apply here even if ESA had types for oauth2 structures, since we made our own custom authenticator). So i still think that if/when we get types from ESA, we would and should be still be directly extending/typing that one interface at the module level to be what really is, instead of it "starting" as oauth2.

pep/app/services/pep-session.ts
```diff
@@ -0,0 +1,27 @@
...
+}) {
+    // normal class body definition here
+}
+
+// DO NOT DELETE: this is how TypeScript knows how to look up your services.
+declare module '@ember/service' {
+    interface Registry {
+        'pep-session': Session;
+    }
+}
```
billdami, line 27: we should rename this class to `PepSessionService` and import it by the same name elsewhere, just to be consistent with all the other service's "Session" suffix and matching the filename to the class name.

I can do this in my branch tho once I merge this into mine, if you want. as I already will have to fix a few conflicts I see.

PR #10, opened 2020-08-28
Adding in query param hook

bakerac4:
Adding in the query param "hook". This allows us to manually control the query params and when they update i.e. on button click. 

The downside here is that we are using a private property on the controller to update the query params. This property seems to tell both the route and controller that things were updated. Otherwise, seems like a decent strategy.

### DIFF
--- a/pep/app/hooks/useQueryParams.ts
+++ b/pep/app/hooks/useQueryParams.ts
@@ -1,0 +1,93 @@
+import { getOwner, setOwner } from '@ember/application';
+import { tracked } from '@glimmer/tracking';
+import DS from 'ember-data';
+import Controller from '@ember/controller';
+
+import { QueryParamsObj } from '@gavant/ember-pagination/utils/query-params';
+import Ember from 'ember';
+
+export type RecordArrayWithMeta<T> = DS.AdapterPopulatedRecordArray<T> & { meta: any };
+
+export interface ResponseMetadata {
+    totalCount: number;
+}
+
+export interface Sorting {
+    valuePath: string;
+    sortPath?: string;
+    isAscending: boolean;
+}
+
+export interface PaginationConfigs {
+    limit?: number;
+    filterList?: string[];
+    includeList?: string[];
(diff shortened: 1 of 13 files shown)


pep/app/hooks/useQueryParams.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Creates and returns a new QueryParams instance and binds its owner to be the same as
+ * that of its parent Controller.
+ * In most cases, this returned instance should be assigned to a @tracked property
+ * on its parent context, so that it can be accessed on the associated template
+ *
+ * If using an object as a query param (eg. Model, POJO), be sure to add a computed to the getter
+ * that grabs the value you need from the object
+ * @param {PaginationArgs} args
```
billdami, line 53: should be `QueryParamArgs`

pep/app/pods/most-cited/route.ts
```diff
@@ -13,14 +16,15 @@ export default class MostCited extends PageNav(Route) {
      * @returns {Promise<Document[]>}
      * @memberof MostCited
      */
-    async model() {
-        const queryParams = buildQueryParams({
+    async model(queryParams: SearchParams) {
```
billdami, line 19: you should probably create a new interface for this page's specific params instead of using the search/route's params interface, as I don't think they actually match up.

pep/app/pods/most-viewed/route.ts
```diff
@@ -3,21 +3,23 @@ import usePagination, { RecordArrayWithMeta } from '@gavant/ember-pagination/hoo
...
+import { useQueryParams } from 'pep/hooks/useQueryParams';
 
 export default class MostViewed extends PageNav(Route) {
     navController = 'most-viewed';
     /**
      * Load the widget results data
      */
-    async model() {
-        const queryParams = buildQueryParams({
+    async model(queryParams: any[]) {
```
billdami, line 14: would make an actual interface for these query params too

PR #11, opened 2020-08-28
Story/tp65873 Search left panel - round 1

billdami:


billdami:
@bakerac4 im not done w/this yet, i still have to change some of the logic for how search queries are generated, but I think its close enough that you can start looking at it, as theres already quite a bit there.

pep/app/pods/search/controller.ts
```diff
@@ -27,15 +38,32 @@ export default class Search extends Controller {
...
+    @tracked viewedCount: string = '';
+    @tracked viewedPeriod: ViewPeriod = SEARCH_DEFAULT_VIEW_PERIOD;
     @tracked paginator!: Pagination<Document>;
 
     @tracked currentSmartSearchTerm: string = '';
     @tracked currentSearchTerms: SearchTermValue[] = [];
     @tracked currentMatchSynonyms: boolean = false;
+    @tracked currentCitedCount: string = '';
+    @tracked currentViewedCount: string = '';
+    @tracked currentViewedPeriod: ViewPeriod = SEARCH_DEFAULT_VIEW_PERIOD;
```
billdami, line 64: note for review (and my future self), going to work on converting this to the new query params hook in the next search PR, as it'll likely require a decent amount of refactoring of all the search stuff, and can't hold this initial PR up any longer.

pep/app/styles/_variables-theme.scss
```diff
@@ -30,3 +24,13 @@ $headings-font-family: Muli, -apple-system, BlinkMacSystemFont, 'Segoe UI', Robo
 $hr-border-color: rgba(#000, 0.06);
 $dropdown-border-color: #dbdbdb;
 $dropdown-border-radius: 0.21rem;
+// ---------------------------
+// Addon overrides
+// ---------------------------
+$button-spinner-loading-color1: rgba(#fff, 0.75);
```
bakerac4, line 30: These seem like the same thing that was put into `tomato`. So Im assuming you can either delete one of these, or you would need to add these to all the other themes

pep/app/utils/search.ts
```diff
@@ -1,57 +1,111 @@
...
+
+            if (searchType.solrField) {
+                const isQuoted = QUOTED_VALUE_REGEX.test(value);
+                const hasMultipleWords = value.indexOf(' ') !== -1;
+                const formatted = hasMultipleWords && !isQuoted ? `"${value}"~25` : value;
+                value = `${searchType.solrField}:(${formatted})`;
+            }
+
+            //if a term of this type already exists, join it to the existing one
+            queryParams[p] = `${qp ? `${qp} ${joinOp} ` : ''}${value}`;
```
bakerac4, line 95: Seems like you could create a small method for this since its used in multiple places

pep/app/utils/search.ts
```diff
@@ -72,16 +126,50 @@ export function buildSearchQueryParams(
...
+/**
+ * A "blank" search request could have the following params:
+ * synonyms, facetfields, and abstract, citecount, viewcount, viewperiod
+ * If there are params besides that, it is a valid search form submission
+ * @export
+ * @param {QueryParamsObj} params
+ * @returns
+ */
+export function hasSearchQuery(params: QueryParamsObj) {
+    const exclude = ['synonyms', 'facetfields', 'abstract', 'citecount', 'viewcount', 'viewperiod'];
```
bakerac4, line 154: You might want to make it so you can pass in the exclude params and just default them that way it can be changed if it they need it to be down the line

pep/app/pods/components/search/form/template.hbs
```diff
@@ -54,9 +54,67 @@
...
+                                type="number"
+                                class="form-control form-control-sm form-control-inline-xs form-control-number-no-spin mx-1"
+                                maxlength="255"
+                                autocomplete="off"
+                                {{on "input" this.onLimitTextChange}}
+                            />
+                            <span class="font-size-secondary text-nowrap">
+                                {{t "search.form.additional.viewed.times"}}
+                            </span>
+                            <select
```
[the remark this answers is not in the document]
billdami: @bakerac4 yeah, the ui for these are not using floating labels, are also using bootstrap's "custom select" which has substantially different styling than the normal form-control/FLselects do. so just going with basic native `<select>` was the cleanest choice for these.

pep/app/constants/search.ts
```diff
@@ -29,6 +29,50 @@ export enum ViewPeriod {
     PAST_12_MONTHS = 4
 }
 
+export enum FacetId {
+    art_id = 'art_id',
```
bakerac4, line 33: @billdami normally I believe we capitalize the left side i.e. ART_ID to make it more clear its a enum/constant

PR #12, opened 2020-08-31
Updates to about modal

### LINKED ISSUE #4: About
It would be helpful to prioritize the "About" functionality of the user menu so we can see what Client release is running on pep-web.rocks, as well as for tagging any issues as associated with a release.  Note that you can also easily get the other component versions via the API endpoint /v2/server/status which returns:
{
  "db_server_ok": true,
  "text_server_ok": true,
  "text_server_version": "8.4.1",
  "opas_version": "2020.0805.1.Alpha",
  "dataSource": "OPAS.AWSProd.2020-08-15",
  "timeStamp": "2020-08-26T20:14:35Z",
  "user_ip": "172.17.0.1"
}

As documented, it's also helpful to show those to understand what's running behind the client as well.

text_server_version is of course the Solr version.  dataSource helps me verify which config is being used, but that's not likely to be useful in your client.

bakerac4:
Fixes #4 . Updates to about modal to provide status, and descriptions of PEP Web and PEP Archive.

### DIFF
--- a/pep/app/pods/components/modal-dialogs/user/about/component.ts
+++ b/pep/app/pods/components/modal-dialogs/user/about/component.ts
@@ -14,6 +14,13 @@ interface ModalDialogsUserAboutArgs {
 export default class ModalDialogsUserAbout extends Component<ModalDialogsUserAboutArgs> {
     @service intl!: IntlService;
 
+    /**
+     * @remarks
+     * This method is part of the {@link ModalDialogsUserAbout}.
+     *
+     * @readonly
+     * @memberof ModalDialogsUserAbout
+     */
     get isStatusOk() {
         return this.args.options.serverInformation.db_server_ok && this.args.options.serverInformation.text_server_ok;
     }
--- a/pep/app/pods/components/modal-dialogs/user/about/component.ts
+++ b/pep/app/pods/components/modal-dialogs/user/about/component.ts
@@ -21,15 +21,12 @@ export default class ModalDialogsUserAbout extends Component<ModalDialogsUserAbo
      * @readonly
      * @memberof ModalDialogsUserAbout
      */
-    get isStatusOk() {
-        return this.args.options.serverInformation.db_server_ok && this.args.options.serverInformation.text_server_ok;
+    get isStatusOk(): boolean {
+        const serverInformation = this.args.options.serverInformation;
+        return serverInformation.db_server_ok && serverInformation.text_server_ok;
     }
 
-    get status() {
-        if (this.isStatusOk) {
-            return this.intl.t('about.status.ok');
-        } else {
-            return this.intl.t('about.status.down');
-        }
+    get status(): string {
+        return this.intl.t(this.isStatusOk ? 'about.status.ok' : 'about.status.down');
     }
 }
--- a/pep/app/pods/components/modal-dialogs/user/about/component.ts
+++ b/pep/app/pods/components/modal-dialogs/user/about/component.ts
@@ -1,12 +1,28 @@
 import Component from '@glimmer/component';
 import { ServerStatus } from 'pep/pods/application/controller';
+import IntlService from 'ember-intl/services/intl';
+import { inject as service } from '@ember/service';
 
 interface ModalDialogsUserAboutArgs {
     onClose: () => void;
     options: {
-        serverInformation: Promise<ServerStatus>;
+        serverInformation: ServerStatus;
         clientBuildVersion: string;
     };
 }
 
-export default class ModalDialogsUserAbout extends Component<ModalDialogsUserAboutArgs> {}
+export default class ModalDialogsUserAbout extends Component<ModalDialogsUserAboutArgs> {
+    @service intl!: IntlService;
+
+    get isStatusOk() {
+        return this.args.options.serverInformation.db_server_ok && this.args.options.serverInformation.text_server_ok;
+    }
+
+    get status() {
+        if (this.isStatusOk) {
+            return this.intl.t('about.status.ok');
+        } else {
+            return this.intl.t('about.status.down');
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


pep/app/pods/components/modal-dialogs/user/about/component.ts
```diff
@@ -1,12 +1,28 @@
...
+    }
+
+    get status() {
+        if (this.isStatusOk) {
+            return this.intl.t('about.status.ok');
+        } else {
+            return this.intl.t('about.status.down');
+        }
+    }
+}
```
billdami, line 28: would condense some of this code:
```ts
get isStatusOk() {
    const info = this.args.options.serverInformation;
    //...
}
```
`status` would probably be cleaner directly in the template:
```hbs
{{t (if this.isStatusOK "about.status.ok" "about.status.down")}}
```

pep/app/pods/components/modal-dialogs/user/about/template.hbs
```diff
@@ -2,14 +2,34 @@
...
     <Modal.body>
         <div class="mb-2">
-            <h5 class="mb-0">{{t "about.client"}}</h5>
-            <p class="text-muted small">{{@options.clientBuildVersion}}</p>
+            <h5 class="mb-0">
+                {{t "about.status.title"}}
+                {{#if this.isStatusOk}}
+                    <FaIcon @icon="check-circle" @prefix="fal" class="text-success"  />
+                {{else}}
+                    <FaIcon @icon="exclamation-circle" @prefix="fal" class="text-danger" />
```
billdami, line 10: I dont think you need `@prefix="fal"` on any of these icons, as thats the default prefix.

pep/app/pods/components/modal-dialogs/user/about/template.hbs
```diff
@@ -22,3 +42,12 @@
...
 </ModalDialog>
+
+
+{{!-- dataSource: "OPAS.AWSProd.2020-08-27"
+db_server_ok: true
+opas_version: "2020.0826.1.Alpha"
+text_server_ok: true
+text_server_version: "8.4.1"
+timeStamp: "2020-08-31T14:23:13Z"
+user_ip: "172.17.0.1" --}}
```
billdami, line 53: do we need this data in the template comments

pep/app/pods/components/modal-dialogs/user/about/component.ts
bakerac4: Im going to disagree with this just based on typescript alone. In general since Ember still doesn't have typed templates, I think that when we can have logic in the classes, we should because that's the only way we get type safety. And I very much doubt they will have typed templates anytime soon. Thoughts?

billdami: I would agree for logic that is more complex yeah, but in a case like this where the logic is so trivial I would actually disagree, especially since this is also purely presentational/formatting logic. You're adding quite a bit of boilerplate (importing a service, injecting it, creating a getter) for something Ember really pushes you to put directly in template as its "happy path" with its various helpers and whatnot.

I think if we continued to go down this rabbit hole of "getters-for-all-the-things" consistently for any logic that possibly could go in a getter, its gonna make things really bloated and probably harder to read/follow just to achieve that extra bit of type safety. This dialog is a very simple UI so I think it seems fine in isolation, but if we followed this same pattern on some more of the complex pages for all the presentational logic/conditionals, we could potentially be adding hundreds of lines of code to class files.

billdami: another thing to consider (that relates to that performance/JS payload size/parse times stuff we were talking about in the ui room the other day), is that generally I think its better to have bigger templates than JS modules in ember-land, when you have a choice between one or the other, in some cases.

Reason being, in prod builds, glimmer compiles templates down to super small and efficient "byte code", which in addition to being small, also have the benefit of not being parsed by the JS runtime upfront when the app first loads, it only gets parsed when its actually used/rendered by the glimmer VM.

bakerac4: I would tend to agree that embers "happy path" would be through the template if they had first class typescript support. The boilerplate - while true ..  we are adding a small amount of it - could be even further decreased to just the getter if we return the keys and do the translation in the template (though its quite minimal even now). As for harder to follow and read, you might be right in a way (in an ideal world we wouldn't even be having this discussion as we would have typed templates) however with the vscode plugins we use and the fact that you can right click on `this.isStatusOk` and it brings you to the class at that line I wouldn't say its hard to follow. I think for the majority of items we would see a benefit and I doubt that we would be adding hundreds of lines very often if at all. 

The performance when we start using engines or async loading of files (which is what much of the web is moving to) should mitigate much of your concern and allow for the same sort of performance increases.

I maintain that the benefits we get from having type safety, outweighs the negatives you have listed. Partially because there are so many things we have not even unlocked with typescript - like generating project documentation based on jsdoc/tsdoc comments.

billdami: Definitely agree that typed templates would be the end-all solution here 💯, and if you had type-safety/autocompletion, etc, you'd see even more logic and presentational logic living in them than we do now.

I can guarantee though, if i were to follow this pattern across the board right now and refactor all the existing code here or in other recent projects, it would add _a lot_ of extra code and boilerplate. Granted, we could mitigate some of this with more DRY code, reusable decorators (similar to the "macros" pattern we used to use in classic ember).

I'll concede in this instance, but just knowing that this pattern won't be universal across in all instances in the project. 😛 And I still think at the very least we should DRY up the code for both these getters (and wherever else we do this) to be as concise/dry as we can:
```ts
get isStatusOk(): boolean {
    const info = this.args.options.serverInformation;
    return info.db_server_ok && info.text_server_ok;
}

get status(): string {
    return this.intl.t(this.isStatusOk ? 'about.status.ok' : 'about.status.down');
}

```

PR #16, opened 2020-09-01
Right Panel Part 2

bakerac4:


billdami:
@bakerac4 that autogenerated typedoc docs adds a ton of files/bloat to the repo, could we add that `pep/documentation` dir to the `.gitignore` for now? 

Since its built/generated it seems like it probably shouldnt be source-controlled, and devs just generate it locally as needed. And/or optimally it'd be nice if we could eventually work this into our build pipeline, where on cand/stage builds it builds the docs and exposes it on some url, similar to swagger on the api side.

bakerac4:
@billdami the reason I added them was I think we can make it public using gh-pages. So that we can link to it and have live documentation through GitHub

billdami:
> @billdami the reason I added them was I think we can make it public using gh-pages. So that we can link to it and have live documentation through GitHub

yeah that would def be nice, though you can have gh-pages do a dynamic build step for stuff (or use something like github Actions to do it), i do that on my personal site. Also, doing it manually in source-control relies on devs remembering to build/commit it as stuff changes so it could easily become outdated, just really feels like the build pipeline should be responsible for this.

billdami:
We could actually even have our Gitlab build pipeline do this, even more easily, since its already mirroring the repo and responsible for builds, and just have it push the compiled docs to the `gh-pages` branch for us, that would definitely be doable.

billdami:
> We could actually even have our Gitlab build pipeline do this, even more easily, since its already mirroring the repo and responsible for builds, and just have it push the compiled docs to the `gh-pages` branch for us, that would definitely be doable.

@bakerac4 also, a good first step towards this ^ would be to do something similar to what ember-cli-release's docs function does, i.e. just put a yarn command in the package.json, e.g. `yarn build:docs` which is just a bash script does something like:
1. `git checkout gh-pages`
2. `your typedoc build command here...`
3. `git add -A` `git commit -m 'generating docs'`
4. `git push -u origin gh-pages`
5. `git checkout master`

And for now we can just run that ourselves until we have gitlab runners doing it for us.

pep/app/pods/components/search/item/component.ts
```diff
@@ -0,0 +1,81 @@
...
+    /**
+     * Using a computed here so we A) dont dip into the local storage too often and B) so that this
+     * is recomputed whenever the preferences object is updated
+     *
+     * @readonly
+     * @memberof SearchItem
+     */
+    @computed('currentUser.preferences', 'args.item')
+    get favorited() {
+        return this.currentUser.hasDocument(PreferenceKey.FAVORITES, this.args.item);
```
billdami, line 28: switching to ID-only storage, this would become a simple `array.includes()` check.

pep/app/pods/components/search/item/component.ts
```diff
@@ -0,0 +1,81 @@
...
+     * @param {PreferenceDocumentsKey} key
+     * @param {Document} document
+     * @memberof SearchItem
+     */
+    toggleDocument(key: PreferenceDocumentsKey, document: Document) {
+        if (this.currentUser.hasDocument(key, document)) {
+            this.currentUser.removeDocument(key, document);
+        } else {
+            this.currentUser.addDocument(key, document);
+        }
```
billdami, line 79: this would be simple immutable update logic switching to ID-only storage

pep/app/pods/components/search/item/template.hbs
```diff
@@ -0,0 +1,45 @@
...
+                {{#each @item.kwicList as |searchHit|}}
+                    <li>{{html-safe searchHit}}</li>
+                {{/each}}
+            </ul>
+        {{/if}}
+    </div>
+    <div class="pl-2 d-flex justify-content-end align-items-start">
+        {{#if @item.relatedrx}}
+            <span class="mr-2">
+                <FaIcon @icon="books" class="p-0 "/>
```
billdami, line 17: seems like the right side might be getting overly icon heavy with all 3 icons (related docs, favorites, read later) aligned to the right. It might be nice to move this related documents indicator over to the left, inline and directly to the right of the documentRef text. Also since its its not an action/button, it would better distinguish it from favorite/read later.

pep/app/pods/components/search/item/template.hbs
```diff
@@ -0,0 +1,45 @@
...
+                    @text={{t "search.item.relatedDocuments"}}
+                    @popperContainer="body"
+                    @side="left"
+                    @spacing={{5}}
+                    @effect="slide"
+                    @innerClass=""
+                />
+            </span>
+        {{/if}}
+        <Button
```
billdami, line 28: these probably should use `Button:WithTooltip` (e.g. "Add to favorites" / "Remove from favorites" for accessibility)

pep/app/pods/glossary-term/model.ts
```diff
@@ -0,0 +1,106 @@
+import DS from 'ember-data';
+import attr from 'ember-data/attr';
+import { belongsTo } from 'ember-data/relationships';
+
+import { isEmpty } from '@ember/utils';
+
+import { INVALID_ABSTRACT_TAGS, INVALID_ABSTRACT_PREVIEW_TAGS, HTML_BODY_REGEX } from 'pep/constants/regex';
+import SimilarityMatch from 'pep/pods/similarity-match/model';
+
+export default class GlossaryTerm extends DS.Model {
```
billdami, line 10: unless its diverged too greatly to make it worth it, this should probably just extend the `Document` model, as it looks almost identical.

pep/app/pods/index/controller.ts
```diff
@@ -20,7 +20,7 @@ export default class Index extends Controller {
     }
 
     get expertPick() {
-        return this.configuration.base.home.expertPick;
+        return this.configuration.base.home.expertPicks[0];
```
billdami, line 23: this should _probably_ be `expertPicks[expertPicks.length - 1]` to show the last one in the list, as i believe they want to show the most recently added one here, but i could be wrong on that.

pep/app/constants/configuration.ts
```diff
@@ -1,5 +1,29 @@
...
+    TWENTY_ONE_BY_NINE = '21by9',
+    SIXTEEN_BY_NINE = '16by9',
+    FOUR_BY_THREE = '4by3'
+}
+
+export interface Publisher {
+    sourceCode: string;
+    previewHTML: string;
+    fullHTML: string;
+}
```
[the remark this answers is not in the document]
bakerac4: The other ones are straightforward but I can if you want

pep/app/pods/components/page/sidebar/widgets/component.ts
```diff
@@ -1,17 +1,21 @@
...
+}
+export interface PageSidebarWidgetArgs extends PageSidebarWidgetsArgs {
     toggleIsOpen: (widget: WIDGET) => void;
     openWidgets: WIDGET[];
 }
 
 export default class PageSidebarWidgets extends Component<PageSidebarWidgetsArgs> {
-    @tracked openWidgets: WIDGET[] = [WIDGET.WHATS_NEW, WIDGET.MOST_CITED, WIDGET.MOST_VIEWED];
+    @tracked openWidgets: WIDGET[] =
+        this.args.widgets?.filter((widget) => widget.open).map((widget) => widget.widget) ?? [];
```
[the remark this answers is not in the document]
bakerac4: Can you explain your though process behind this?

pep/app/pods/components/page/sidebar/widgets/more-like-these/component.ts
```diff
@@ -1,18 +1,68 @@
...
+            const results = yield this.store.findRecord('document', this.data.id, {
+                reload: true,
+                adapterOptions: { query: { similarcount: this.similarCount } }
+            });
+            const matches = results.similarityMatch.similarDocuments.filter(
+                (item: Document) => item.id !== this.data.id
+            );
+
+            // TODO Right now the document that the find was called for is somehow being added into
+            // the array. We should figure out why and fix it
```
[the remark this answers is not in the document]
bakerac4: Yea there is a good chance it is, I just didn't have any more time to fool around with it. I was seeing the similar docs include the document that we are getting the similar docs for.

pep/app/pods/components/page/sidebar/widgets/component.ts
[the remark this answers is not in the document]
billdami: its mainly that defaulting class properties using values from instance properties (usually `args`) is always shown as done in the `constructor()` in ember docs/examples, and its a little easier for someone unfamiliar to see that this is just a one-time "setup" type logic on initial render if its done there, rather than possibly being some sort of auto-updated value that will always reflect the current `args` values.

I'm pretty sure I had also seen some potential gotchas around using this pattern at one point, though I don't remember what now.

pep/app/pods/components/page/sidebar/widgets/more-like-these/component.ts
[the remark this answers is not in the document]
billdami: hm yeah, it might be that you need to use some of the relationship configs, like `inverse: null` or something along those lines to tell ember-data how to treat the relationship. 

https://guides.emberjs.com/release/models/relationships/#toc_explicit-inverses

I noticed you had some pretty funky stuff going on in the serializer too, so thats another place i'd probably examine more closely too if i were to look at it.

pep/app/constants/preferences.ts
```diff
@@ -2,30 +2,39 @@ import ENV from 'pep/config/environment';
...
 }
 
 export interface UserPreferences {
+    favorites?: Document[];
     preferencesVersion: string;
     theme: ThemeId;
     lang: LanguageCode;
     searchLimitIsShown?: boolean;
     searchTermFields?: SearchTermId[];
+    readLater?: Document[];
```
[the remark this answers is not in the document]
bakerac4: While I agree with some of your points, I don't like the fact that we would have to keep making API calls to get the same data - over and over and over again. I could possibly compromise on loading id, and the text we need to display the items instead of the whole object.

pep/app/pods/application/controller.ts
```diff
@@ -38,6 +40,9 @@ export default class Application extends Controller {
     @tracked viewedPeriod: ViewPeriod = ViewPeriod.PAST_WEEK;
     @tracked searchTerms: SearchTermValue[] = [];
 
+    @tracked rightSidebarWidgets = this.configuration.base.global.cards.right;
+    @tracked leftSidebarWidgets = this.configuration.base.global.cards.left;
```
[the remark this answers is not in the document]
bakerac4: I'd prefer to do something like

```
@tracked readonly rightSidebarWidgets = this.configuration.base.global.cards.right;
@tracked readonly leftSidebarWidgets = this.configuration.base.global.cards.left;
```

Not only is it more compact, but it also looks cleaner too. It also still doesn't allow you to overwrite it

pep/app/constants/preferences.ts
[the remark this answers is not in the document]
billdami: I agree, that if you say had 50 favorites and added a new favorite, it wouldnt be great to have to do an api call that reloads ALL 51 of those documents. We could make it smart tho w/o too much effort, by when their favorites update/change, using ember-data's cache and peeking for documents that already exist in the cache (which in this case would be all of them), and then only doing an api request for those that are not cached (which would really only happen on initial card load).

If we reaallly didnt want to try that though, at the very least yeah we should only be storing a "partial" document with the bare minimum fields needed to render the card (ID, title, author, etc). But knowning that the caveat here is we'd have to make concessions/logic in the code if they ever wanted to use additional fields here after release.

Also, does the reqs doc mention an upper limit anywhere for things like the user's number of Read Later or Favorite items? If not, we should probably discuss that too, as at the very least it should probably have some sane hard cap, not only for storage concerns but rendering concerns too.

pep/app/pods/components/page/sidebar/widgets/component.ts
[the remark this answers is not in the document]
bakerac4: So Im not sure how there could be gotchas as this is just syntactic sugar for setting up properties in the constructor. Also if they are familiar with es6 classes at all, they really shouldn't be confused. If you find out if there really are gotchas I'll concede but otherwise I think this is cleaner and more straightforward.

pep/app/pods/application/controller.ts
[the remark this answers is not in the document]
billdami: I think you are misusing `@tracked` here though. `@tracked` is doing a bunch of stuff behind the scenes to tell ember/glimmer to observe changes to it and all that jazz, which is redundant and not needed here, so while the code is more compact, its actually making the app/resource usage heavier. Using getters for this is also already an established pattern throughout the app, so its not great that we'd have 2 patterns for the same thing now.

We can possibly achieve a similar compact/clean pattern like that tho by creating a custom "alias" decorator, which is really what you are after here, that would be alot more lightweight. We did something similar in some of our addons with an `@argDefault('...')` decorator, but basically it could look like this:

```ts
@alias('configuration.base.global.cards.left') readonly rightSidebarWidgets: Widgets[];
```
and internally, the decorator would basically do exactly what the usual getter pattern does.

pep/app/pods/components/page/sidebar/widgets/component.ts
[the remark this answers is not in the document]
billdami: I thought it had something to do with the rendering lifecycle and how/when `args` are passed in, that could sometimes lead to hard to debug rendering quirks when you do something like this. I've rarely if ever seen `this.args` or even just the `this` context used to default class fields in their class body declaration, its always in the constructor.

I'll see if I can find more info on that though. So if you really don't want to change it, I would just keep an eye on it at least for any weirdness.

pep/app/pods/application/controller.ts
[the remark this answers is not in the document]
bakerac4: From what I remember, a getter is just auto tracked. So the only different would be manually telling it to be tracked?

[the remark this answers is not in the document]
billdami: I dont think so, getters just auto-recalculate whenever one of the properties they use change which is just ES6 class functionality, but in ember-land, if the updated properties used by the gettter are not `@tracked`, the getter will still recalc, but it won't trigger a re-render, iirc.

By explicitly adding `@tracked` here, you are essentially creating a duplicate tracking observer for a property, `this.configuration.base`, that is already tracked.

Also, I think there might be another issue w/this pattern that I just realized. Here, you are just defaulting/initializing the `rightSidebarWidgets` field with the current value of `this.configuration.base.global.cards.right` on initial render. But if that configuration were to later change (e.g. if the user changes their language or somthing, we reload all configs from the API), it would not have the updated value. In this case, its probably unlikely an action like that would change these specific admin configs, but we should probably avoid that to be safe, and to set the pattern for other configs that may change.

pep/app/pods/glossary-term/model.ts
bakerac4: Yea I think there are fields that are not in documents. I think eventually we should create a base model to inherit from but If you don't mind we can save that for a later date

billdami: can we just extend from `Document` for now and then just add whatever fields u are using that are not in the Document model?

pep/app/pods/application/controller.ts
```diff
@@ -39,9 +38,8 @@ export default class Application extends Controller {
     @tracked viewedCount: string = '';
     @tracked viewedPeriod: ViewPeriod = ViewPeriod.PAST_WEEK;
     @tracked searchTerms: SearchTermValue[] = [];
-
-    @tracked rightSidebarWidgets = this.configuration.base.global.cards.right;
-    @tracked leftSidebarWidgets = this.configuration.base.global.cards.left;
+    @tracked readonly rightSidebarWidgets = this.configuration.base.global.cards.right;
+    @tracked readonly leftSidebarWidgets = this.configuration.base.global.cards.left;
```
billdami, line 42: I still think these need to be getters (or use a new `@alias` decorator that has the same behavior as the getter) because of the concerns we talked about (redundant `@tracked`, not consistent w/the existing pattern). But mainly because of the larger issue that these won't update if `this.configuration.base` updates later.

pep/app/pods/components/page/sidebar/left/content/component.ts
```diff
@@ -1,13 +1,13 @@
...
+import { inject as service } from '@ember/service';
 import { WidgetData } from 'pep/constants/sidebar';
 interface PageSidebarLeftContentArgs {
     data: WidgetData;
 }
 
 export default class PageSidebarLeftContent extends Component<PageSidebarLeftContentArgs> {
-    @inject configuration!: ConfigurationService;
+    @service configuration!: ConfigurationService;
     @tracked leftSidebarWidgets = this.configuration.base.global.cards.left;
```
billdami, line 12: same here as the comment on application/controller.ts

pep/app/pods/components/page/sidebar/right/component.ts
```diff
@@ -1,13 +1,13 @@
...
 import { WidgetData } from 'pep/constants/sidebar';
 interface PageSidebarRightArgs {
     data: WidgetData;
 }
 
 export default class PageSidebarRight extends Component<PageSidebarRightArgs> {
-    @inject configuration!: ConfigurationService;
-    @tracked rightSidebarWidgets = this.configuration.base.global.cards.right;
+    @service configuration!: ConfigurationService;
+    @tracked readonly rightSidebarWidgets = this.configuration.base.global.cards.right;
```
billdami, line 12: same here as the comment on application/controller.ts

pep/app/pods/components/page/sidebar/widgets/favorites/component.ts
```diff
@@ -36,7 +55,7 @@ export default class PageSidebarWidgetsFavorites extends Component<PageSidebarWi
      */
     @action
     onElementChange() {
```
billdami, line 57: missed this before, but we should rename this to be more accurate, maybe `onPreferencesUpdate()`

pep/app/pods/components/page/sidebar/widgets/favorites/component.ts
```diff
@@ -25,8 +32,20 @@ export default class PageSidebarWidgetsFavorites extends Component<PageSidebarWi
      *
      * @memberof PageSidebarWidgetsFavorites
      */
-    loadFromUserPreferences() {
-        this.results = this.currentUser.getDocuments(PreferenceKey.FAVORITES);
+    @restartableTask
+    *loadFromUserPreferences() {
+        const ids = this.currentUser.getDocuments(PreferenceKey.FAVORITES);
```
billdami, line 37: could we put a `TODO` comment here about adding the smart pulling of documents from the local cache?

pep/app/pods/components/page/sidebar/widgets/favorites/component.ts
```diff
@@ -36,7 +55,7 @@ export default class PageSidebarWidgetsFavorites extends Component<PageSidebarWi
      */
     @action
     onElementChange() {
-        this.loadFromUserPreferences();
+        taskFor(this.loadFromUserPreferences).perform();
```
billdami, line 58: This also should have some logic so that it only refetches the documents if the array of document ids changed, not just anytime prefs change, since we wouldnt want this to reload if the user changed their Theme for example. 

This would possibly be less necessary if we had the smart cache usage, tho probably still good to have, but maybe it can just be a `TODO` comment for now?

pep/app/pods/components/page/sidebar/widgets/read-later/component.ts
```diff
@@ -25,16 +32,28 @@ export default class PageSidebarWidgetsReadLater extends Component<PageSidebarWi
...
+            const results = yield this.store.query('document', params);
+            this.results = results.toArray();
+        }
     }
 
     /**
      * Load the widget results on render
      */
     @action
     onElementChange() {
```
billdami, line 55: same here as the favorites component regarding renaming this method, e.g. `onPreferencesUpdate()`

pep/app/pods/components/page/sidebar/widgets/related-documents/component.ts
```diff
@@ -55,14 +55,14 @@ export default class PageSidebarWidgetsRelatedDocuments extends Component<PageSi
     @dontRunInFastboot
     onElementInsert() {
         if (this.data?.relatedrx) {
-            taskFor(this.loadResults).perform();
+            taskFor(this.loadResults).perform(this.data?.relatedrx);
         }
     }
 
     @action
     onElementChange() {
```
billdami, line 63: this one too, `onDataUpdate()`

pep/app/pods/components/search/item/bibliographic/template.hbs
```diff
@@ -12,25 +25,14 @@
...
-                    @text={{t "search.item.relatedDocuments"}}
-                    @popperContainer="body"
-                    @side="left"
-                    @spacing={{5}}
-                    @effect="slide"
-                    @innerClass=""
-                />
-            </span>
-        {{/if}}
+        {{!-- TODO: Why does using Button::Tooltip here not update the iconPrefix --}}
```
billdami, line 28: When you use `Button::WithTooltip`, you'll need to update that wrapper component to pass `@iconPrefix` down into the inner `Button` component, i didn't get all of those args when i created it.

pep/app/pods/index/controller.ts
```diff
@@ -20,7 +20,8 @@ export default class Index extends Controller {
     }
 
     get expertPick() {
-        return this.configuration.base.home.expertPicks[0];
+        const expertPicks = this.configuration.base.home.expertPicks;
+        return expertPicks[expertPicks.length - 1];
```
billdami, line 24: make sure to make the same change in the index/route's `model()`

pep/app/services/current-user.ts
```diff
@@ -216,11 +215,11 @@ export default class CurrentUserService extends Service {
      * @param {Document} document
      * @memberof CurrentUserService
      */
-    addDocument(key: PreferenceDocumentsKey, document: Document) {
+    addDocument(key: PreferenceDocumentsKey, documentId: string) {
```
billdami, line 218: at the very least i think these methods should be renamed, e.g. `addPrefDocument()` but I still dont think they really belong in the current-user service at all, especially since updating these can now be a one-liner.

`this.loadLocalStoragePrefs();` should not be necessary since the `currentUser.preferences` should always be up-to-date with what's in localStorage. All this logic could be condensed and done directly in the search item component as:
```ts
this.currentUser.updatePrefs({
    [PreferencesKey.FAVORITES]: [...this.currentUser.preferences[PreferencesKey.FAVORITES], id].uniq()
})
```

billdami (changes requested):

pep/app/pods/components/page/sidebar/widgets/component.ts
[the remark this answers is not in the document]
billdami: Yep, i was wrong on this one 😛 so this should be safe moving forward.

pep/app/pods/components/page/sidebar/widgets/favorites/component.ts
bakerac4: Thats partly correct. This function is both called on `did-insert` and `did-update`. When running on `did-insert` the preferences were not updated

pep/app/services/current-user.ts
bakerac4: I find it cleaner to abstract that out

i.e. 
`this.currentUser.addDocument()`
is a lot cleaner than 
```
this.currentUser.updatePrefs({
    [PreferencesKey.FAVORITES]: [...this.currentUser.preferences[PreferencesKey.FAVORITES], id].uniq()
})
```
and since we add and remove from multiple places I think the way I have it provides a DRY'er experience

bakerac4: But I can change the name

pep/app/pods/components/page/sidebar/widgets/favorites/component.ts
billdami: Yeah I missed that. in that case, maybe `onInsertOrPrefsUpdate()` or even just two separate actions, `onElementInsert()` and `onPreferencesUpdate()` that both call the load method. This might be ideal because if/when we add diff checking for the IDs array, that should only happen in the `did-insert` path.

pep/app/pods/components/search/item/bibliographic/template.hbs
```diff
@@ -25,14 +25,13 @@
         {{/if}}
     </div>
     <div class="pl-2 d-flex justify-content-end align-items-start">
-        {{!-- TODO: Why does using Button::Tooltip here not update the iconPrefix --}}
         <Button
```
billdami, line 28: think u are missing the `::WithTooltip` here and below

pep/app/services/current-user.ts
```diff
@@ -215,7 +215,7 @@ export default class CurrentUserService extends Service {
      * @param {Document} document
      * @memberof CurrentUserService
      */
-    addDocument(key: PreferenceDocumentsKey, documentId: string) {
+    addPreferenceDocument(key: PreferenceDocumentsKey, documentId: string) {
         const prefs = this.loadLocalStoragePrefs();
```
billdami, line 219: these should still not be using `this.loadLocalStoragePrefs();`, it should just use `this.preferences`. same for the other document-related methods below.

pep/translations/en-us.json
```diff
@@ -335,7 +335,7 @@
             },
             "relatedDocuments": {
                 "title": "Related Documents",
-                "noResults": "You must view a document to see the documents related to it"
+                "noResults": "You must view a document to see the documents related to it. If you are already viewing one, this means no results have been found"
```
billdami, line 338: i think it would be cleaner to have this message be conditional. show 
> You must view a document to see the documents related to it.
if there is no related document currently set in the sidebar data, otherwise show
> No related documents found.

But we dont need to hold up the PR for this if you dont want to do it now, think it should be a polish item eventually tho.

bakerac4: FWIW - I had the same conversation in my head

billdami (approved):

PR #18, opened 2020-09-02
Adding in CSV handling

bakerac4:


pep/app/utils/url.ts
```diff
@@ -82,3 +87,17 @@ export function appendTrailingSlash(url: string) {
     const hasQueryStr = url.indexOf('?') !== -1;
     return hasQueryStr ? url.replace('?', '/?') : `${url}/`;
 }
+
+/**
+ * Generate a CSV url for documents
+ *
+ * @param {DS.Store} store
+ * @param {QueryParamsObj} queryParams
+ * @returns
```
billdami, line 96: prob should be `@returns {string}`

Issue #19, opened 2020-09-02
Preliminary Review of Stage per the 4pm  or so build


nrshapiro:
Overall, looking very Good!

I like the way search "preview" in the Refine area works.  

## Two issues, in case you don't know yet:
1) refine is sticky in a bad way
   - I smart searched for Tuckett 59 entries
   - I selected Turkish, which had one entry.
   - Refined and now down to one listing.
   - Cleared search.
   - Smart searched for Levinson
   - no hits (notice the language filter is still in the URL)
   - Smart searched for Tuckett
   - 1 entry found, the Turkish one.

2) A number of Abstract previews displaying weird with a dark box over some of the text.  Sort of like Redacted!

## Other Comments
1) Most Cited and Most Viewed intervals would be better named more plainly ... I suggest simply "# Times Cited", "# Times Viewed"  People will see that there are intervals and certainly get it.

2) In using Refine, I noticed it's fast enough that it might be better to work without the Refine button.  Why not just update the set as I click on the items?  

3) Search speed is generally pretty good, but some searches can be slower than I'm used to seeing--I think that's just a function of running too small a Solr machine or having too little memory for solr or a too small JVM setting, but certainly for demos we'd want a faster instance.

4) For demo, I'd like to remove one of the Glossary refine areas...I knew those would be similar, and I do want to experiment with it more to see which is better.  Right now, they are almost ALWAYs the same, and that will give a negative impression.  (Or put one of them at the end of the list.)  Yes, when we have an admin, we will be able to configure that, but for demos before that, it's best to remove or move one.)  
  ETA:  As I suspected, Glossary Groups has some better terms.  It may not have as many, but I think I'd keep that one and put the other offline for demos now.  Like other refine fields though, this is a much easier way to tune them than I've had before.

I'll play with it more later.  Exciting progress!

## BTW - It's fun to play with smartsearch this way.  (I've been waiting for that.)  I have only seen it in the OpenAPI interface and in my test programs.  Here's a little exercise (which works, but I do see something I might change):
- Tuckett finds papers by Tuckett
- tuckett searches the text for tuckett.  As implemented: it's not a name if it's not capitalized.  I might change that to be smarter if the name is found in the author list.  thinking about it.
- Obama searches the text for Obama, since it knows that Obama isn't one of the authors in the database.
- Tuckett (2015) finds his papers from 2015
- Pasting in 
[...]

PR #20, opened 2020-09-03
Story/tp65873 Search left panel - round 2 (feedback)

### LINKED ISSUE #15: Clearing Refine Settings
Perhaps it's just not in yet, but Clear Search doesn't clear the refine settings.  Perhaps it shouldn't, so you can leave a choice of English, for example, in your search criteria, but then there needs to be an easy way to clear all the refine settings in case you have multiple selections made, and want to start over. 

Perhaps you could do this in like manner to the way the search buttons are laid out.
   `Clear  [REFINE]'

billdami:
Closes #15, #19, and #21


pep/app/pods/search/route.ts
```diff
@@ -116,7 +118,7 @@ export default class Search extends PageNav(Route) {
         controller.currentViewedCount = controller.viewedCount;
         controller.currentViewedPeriod = controller.viewedPeriod;
         controller.currentSearchTerms = isEmpty(controller.searchTerms)
-            ? [{ type: 'everywhere', term: '' }]
+            ? cloneDeep(SEARCH_DEFAULT_TERMS)
```
bakerac4, line 121: My minor gripe with this is I really don't love the name `cloneDeep` because its not clear why or what its doing this for. 

Thoughts about creating a `getInstance` helper that calls `cloneDeep` in it?

billdami: i think `cloneDeep` more accurately describes what its actually doing tho, since its literally deeply/recursively cloning the given object, its not getting an instance of something via like `new obj()` or anything.

I could just rename the import to like `copy` or `copyObject()` or something if you preferred? As well as add comments here as to why its needed.

bakerac4: That sounds good

Issue #24, opened 2020-09-03
Facets Issue and enhancement...


nrshapiro:
1) Are you sure you are using field `art_kwds_str `for the facets?  I searched for projective identification "everywhere".  I don't see it in the keywords list.  That suggests you are faceting the text field  `art_kwds `instead of `art_kwds_str`.  When I do the search in Solr, the facet list includes a number of multi-word keywords (including one ugly one with spaces and a \n in it I'll have to fix.)

   Searching "loving memory"~25 in solr

   ```
     facet art_kwds_str:
        "conflict",2,
        "countertransference",2,
        "implicit relational\n                        knowing",2,
        "interpretation",2,
        "intuition",2,
        "loving self",2,
        "mutual parental love",2,
        "neuroscience",2,
        "absence",1,
        "addictive pleasure",1,
        "adult attachment interview",1,
        "affect regulation",1,
   ```

2) Add field `authors` to the facet list.  This is not a rush, but if easy to add, I think it would be great to have soon.  This was not included in the specified list for facets...but part of my trying to make refine as useful as possible.  

   So the list, in order

   1) Decade
   2) View Periods (another rename suggestion: View Count)
   3) Citation Periods (another rename suggestion: Citation Count)
   4) Author (field `authors`)
   5) Language
   6) Article Type
   7) Source Type
   8) Source
   9) Glossary Groups
   10) Keywords (field `art_kwds_str`)

Top 15 results for  facet authors in Solr for search loving memory"~25:

```
    "Stolorow, Robert D.",5,
        "Auerhahn, Nanette C.",3,
        "Frosh, Stephen",3,
        "Natterson, Joseph",3,
        "Caruth, Elaine G.",2,
        "Freud, Sigmund",2,
        "Harris, Martha",2,
        "Krystal, Henry",2,
        "Laub, Dori",2,
        "Levine, Howard B.",2,
        "Abbass, Allan",1,
        "Alston, Edwin F.",1,
        "Alvarez, Anne",1,
        "Astor, James",1,
        "Bach, Sheldon",1,
```

billdami:
@nrshapiro I did switch the Keywords facet to use `art_kwds_str` in one of the most recent pushes, that I believe that might not be deployed atm (we've been having some deploy/build issues today). 

[...]

PR #26, opened 2020-09-04
Story/tp65873 Search left panel - round 4 (requirements)

### LINKED ISSUE #3: Animation during screen updates
One of the great things about an SPA is that the entire page stays constant, while new data is added/removed/updated.

The full screen animation during update delays really defeats that. 

(I recognize I'm looking at a very early version...but the point of writing these issues is to focus on getting some early "tuning".

billdami:
Closes #3, #24


pep/app/pods/components/search/refine/component.ts
```diff
@@ -30,15 +31,18 @@ interface SearchRefineArgs {
...
+    @service configuration!: ConfigurationService;
 
     @tracked expandedGroups: string[] = [];
 
     get groups() {
+        const cfg = this.configuration.base.search;
         const groups: RefineGroup[] = [];
         const fieldsMap = this.args.metadata?.facetCounts?.facet_fields ?? {};
         const incFields = Object.keys(fieldsMap);
+        const displayedFacets = cfg.facets.defaultFields.map((id) => SEARCH_FACETS.findBy('id', id)!);
```
bakerac4, line 43: I have a feeling using `!` is probably the wrong way to go here

pep/app/pods/most-cited/template.hbs
```diff
@@ -9,12 +9,12 @@
                 }}
             </h4>
             <div class="d-flex align-items-center">
-                <Button
+                <Button::WithTooltip
                     @type="link"
                     @icon="download"
                     @iconSize="lg"
                     @action={{this.downloadCSV}}
-                    title={{t "common.export"}}
```
bakerac4, line 17: How does this work with accessibility? Do you still need title as well?

pep/app/pods/most-viewed/template.hbs
```diff
@@ -9,12 +9,12 @@
...
             </h4>
             <div class="d-flex align-items-center">
-                <Button
+                <Button::WithTooltip
                     @type="link"
                     @icon="download"
                     @iconSize="lg"
                     @action={{this.downloadCSV}}
-                    title={{t "common.export"}}
+                    @tooltip={{t "common.export"}}
```
bakerac4, line 17: Same thing here, does this change accessibility?

pep/app/pods/read/document/controller.ts
```diff
@@ -49,7 +48,7 @@ export default class ReadDocument extends Controller {
     @tracked viewedPeriod: ViewPeriod = SEARCH_DEFAULT_VIEW_PERIOD;
     //workaround for bug w/array-based query param values
     //@see https://github.com/emberjs/ember.js/issues/18981
-    @tracked _searchTerms: string | null = JSON.stringify(SEARCH_DEFAULT_TERMS);
+    @tracked _searchTerms: string | null = null;
```
bakerac4, line 51: Do you need the comment above this anymore?

pep/app/pods/read/document/route.ts
```diff
@@ -38,33 +43,57 @@ export default class ReadDocument extends PageNav(Route) {
...
-                pagingRootKey: null,
-                filterRootKey: null,
-                processQueryParams: (params) => ({ ...params, ...searchParams })
-            });
+            //if no search was submitted, don't fetch any results
+            if (hasSearchQuery(searchParams)) {
+                const controller = this.controllerFor(this.routeName);
+                const queryParams = buildQueryParams({
+                    context: controller,
+                    pagingRootKey: null,
```
bakerac4, line 83: While it doesn't need to be fixed for this PR, I'd like to take a look at why we need to specify `pagingRootKey` and `filterRootKey` as null almost every time we use the pagination. We should just default it so we don't have to worry about it

billdami: `buildQueryParams` is a util function that comes from our ember-pagination addon, which has its defaults optimized for JSON-API, which have the root keys for those as `page` and `filter`, which params get nested under. I guess we could maybe create an "overridden" version of this function in PEP that proxies to the addon's function w/those set to `null`? I would still keep the addon defaults to json-api tho is thats what most of our internal-api based projects will be using moving forward.

pep/app/pods/most-cited/template.hbs
billdami: The addon actually smartly adds an `aria-describedby` attribute on the anchor element (the `<button>` here) that points to the id of the tooltip content container, so we should be good accesibility-wise I think.

pep/app/pods/most-viewed/template.hbs
billdami: see comment on above instance.

pep/app/pods/components/search/refine/component.ts
billdami: yeah so its due to TS not knowing how ember's convenience `findBy()` returns, and that it is guaranteed to return an object here.

The alternative to the `!`, is the more verbose:
```
.map((id) => SEARCH_FACETS.find((facet) => facet.id === id));
```
But i'm fine with switching to that if you prefer (just losing the ability to use all of ember's nice convenience methods that we pay for in our vendor payload size is 😢)

PR #31, opened 2020-09-15
Misc fixes for right cards and tables

bakerac4:


pep/app/pods/most-viewed/controller.ts
```diff
@@ -123,6 +123,7 @@ export default class MostViewed extends Controller {
             filterRootKey: null,
             filterList: ['author', 'title', 'sourcename', 'period', 'queryType']
         });
+        delete queryParams.limit;
```
billdami, line 126: we prob should eventually update that addon's util to allow `limit: null` to do this.

bakerac4: I can do that later on if your ok with that

billdami (approved):

PR #33, opened 2020-09-17
Story/tp65864 Sidebar resizability

billdami:


pep/app/services/current-user.ts
```diff
@@ -199,10 +203,9 @@ export default class CurrentUserService extends Service {
      * @memberof CurrentUserService
      */
     addPreferenceDocument(key: PreferenceDocumentsKey, documentId: string) {
```
billdami, line 205: updated these methods to use pure immutable update logic, I forgot to mention that requirement on the other PR passes.

pep/app/pods/components/page/sidebar/component.ts
```diff
@@ -1,17 +1,58 @@
...
+import { next } from '@ember/runloop';
+import { htmlSafe } from '@ember/template';
 
 import SidebarService from 'pep/services/sidebar';
+import FastbootMediaService from 'pep/services/fastboot-media';
+import { SIDEBAR_HANDLE_WIDTH, SIDEBAR_WIDTH } from 'pep/constants/dimensions';
 
 interface PageSidebarArgs {
     side: 'left' | 'right';
+    resizable?: boolean;
```
bakerac4, line 14: From what I can see, these are not actually being passed in? If that's the case, why are they in the `PageSidebarArgs`?

pep/app/pods/components/page/sidebar/component.ts
```diff
@@ -1,17 +1,58 @@
...
+    }
+
+    /**
+     * Sidebars default to a max width of 35% of the current window width on desktop
+     * and 50% on tablets (as we only allow 1 sidebar open at a time on tablet)
+     * @readonly
+     * @returns {number}
+     */
+    get maxWidth() {
+        const pct = this.fastbootMedia.isTablet ? 0.5 : 0.35;
```
bakerac4, line 43: Bringing these out to a constants file would be nice

pep/app/pods/components/page/sidebar/component.ts
```diff
@@ -29,11 +70,86 @@ export default class PageSidebar extends Component<PageSidebarArgs> {
         return this.isLeft ? this.sidebar.rightSidebarIsOpen : this.sidebar.leftSidebarIsOpen;
     }
 
+    get sidebarStyles() {
+        return this.resizedWidth && this.isOpen
+            ? htmlSafe(`width: ${this.resizedWidth + SIDEBAR_HANDLE_WIDTH}px;`)
+            : null;
```
bakerac4, line 76: I don't actual know the answer without experimenting .. but can you just returned undefined instead of null?

pep/translations/es-es.json
```diff
@@ -67,6 +67,13 @@
             "label": "Sort:",
             "year": "Year"
         },
+        "item": {
```
bakerac4, line 70: I would recommend taking these out so we still get warnings. Otherwise your temporarily solving a build issue that will turn into a runtime issues later on

billdami: not sure i follow. the spanish translations file was just created to test the i10n/i18n switching stuff early on, and I've been keeping it in sync w/the english file, just duplicating the english translations for now, as it will ultimately need the exact same keys duplicated (probably starting off w/the english strings), and then handed off to the PEP people to translate.

I can just disable spanish in the app for now and then just delete this file until we are ready to translate, rather than seeing these housekeeping commits as we go along, if we prefer.

bakerac4: I believe I have observed build errors when you have a key that's in the English file, that's not in the Spanish one. I was leaving those as it acted as a reminder that we need to get them translated. If you want to disable Spanish and delete the file for now, that's fine too as long as we just have a story to remind ourselves to get all the translations

pep/app/pods/components/page/sidebar/component.ts
billdami: probably should work, but we typically always pass `null` to `style=` attributes in ember when we dont want any styles rendered, any reason you prefer `undefined` here instead?

bakerac4: It would allow you make this code more concise, transforming it to

```
get sidebarStyles() {
    return this.resizedWidth && this.isOpen && htmlSafe(`width: ${this.resizedWidth + SIDEBAR_HANDLE_WIDTH}px;`)
}
```

pep/app/pods/components/page/sidebar/component.ts
billdami: they are optional right now and are used in the getters below, but just added some basic knobs here in case resize-ablity needs to be customizable later.

pep/app/pods/components/page/sidebar/component.ts
billdami: in that example, if `resizedWidth` is zero, it would return zero, or if `isOpen` is false, it would return false, so I'd feel less confident about returning values like that to the style attribute.

pep/translations/es-es.json
billdami: yea theres a separate story/release altogether to actually enabling language switching/translations (I believe its not even for the first release, Rel.2 i think).

But yeah, I think the intention is to hand them the json files that contain all the keys/default english translations, so that they can then convert and translate to each language. So having the build warnings for missing translations is just unnecessary noise for us, since we aren't going to be responsible for actually putting in translations ourselves as we add new stuff for the initial release.

billdami:
@bakerac4 this is ready for another pass

pep/app/pods/components/page/sidebar/component.ts
```diff
@@ -1,17 +1,63 @@
...
+    get resizable() {
+        return this.args.resizable ?? true;
+    }
+
+    get minWidth() {
+        return this.args.minWidth ?? SIDEBAR_WIDTH;
+    }
+
+    /**
+     * Sidebars default to a max width of 35% of the current window width on desktop
```
bakerac4, line 42: I wouldn't put the percentage #'s in the comment since they are constants.

PR #35, opened 2020-09-18
Story/tp65735 Updating user prefs

billdami:


pep/app/services/current-user.ts
```diff
@@ -1,10 +1,10 @@
 import Service, { inject as service } from '@ember/service';
+import { assign } from '@ember/polyfills';
```
bakerac4, line 2: @billdami I do not believe you need to use assign from `@ember/polyfills`. It looks like all browsers we support have supported `Object.assign` since 2015

https://caniuse.com/?search=object.assign

pep/app/pods/components/search/form/component.ts
```diff
@@ -57,6 +60,10 @@ export default class SearchForm extends Component<SearchFormArgs> {
         );
     }
 
+    get hasEnteredLimits() {
```
bakerac4, line 63: @billdami It doesn't look like `hasEnteredLimits` is being used anywhere currently. Should this be removed?

billdami:
@bakerac4 ready for another pass

PR #39, opened 2020-09-22
Story/tp65875 Search results preview

### LINKED ISSUE #37: Search by cited and viewed values should be allowed as well as simply limiting search
@billdami 

Separated per our discussion on another issue, cited and viewed search, which currently only work on their own, should be allowed independently as well.  

However, I do (personally) like how they are hidden.  It cleans up the form.  In my view, these fields are not even important to have, because you can use the refine facets to do the same thing after the search.  So perhaps leave it under a button like limit.  

People do like the way PEPEasy hides some of the fields (which the user could configure in options), and then expose it using an expand the form button (in that case a +, but it could have just said "more").  Especially important for mobile use.  Later, if we decide to add other "extra"/minor field types but we don't want to complicate the form, we can change "limit" to "more" and add some additional fields there.  We'll see on the next review to the board what they think.  It should be easy to open that up by default anyway (and then hide it instead if the user so chooses).

They may want them to show, but, again, we'll see.  I'm mainly judging this because they "liked" the old GVPi form, except for the order of the fields.  It's important to at least some of the board that the phrase and search field is on top.  Of course, it isn't, because that's smartsearch, and at some point  I may have you move SmartSearch down, so we can do more interesting things with it (there was already a complaint for the early production site preview that Smartsearch would determine you were better off with a title search instead of a term/phrase search--so I had to turn that feature off.  But it actually was a more precise search than term and phrase search.  I don't want to move it yet, because IMHO, smartsearch is better at the top.  But if it's going to be limited, we might have to move it eventually.

billdami:
Closes #37

### DIFF
--- a/pep/app/pods/components/page/sidebar/widgets/template.hbs
+++ b/pep/app/pods/components/page/sidebar/widgets/template.hbs
@@ -1,20 +1,22 @@
-<div class="px-2 pb-1 d-flex justify-content-between align-items-center" ...attributes>
-    <h5 class="m-0">{{t "relatedInfo.title"}}</h5>
-    <Button
-        @type="link"
-        @action={{this.toggleAll}}
-        @label={{t (if this.showCloseAll "relatedInfo.batchClose" "relatedInfo.batchOpen")}}
-        @icon={{if this.showCloseAll "chevron-up" "chevron-down"}}
-        class="p-0 font-size-secondary"
-    />
-</div>
-<div class="p-2">
-    {{#each @widgets as |widget|}}
-        {{component
-            (concat "page/sidebar/widgets/" widget.widget)
-            toggleIsOpen=this.toggleIsOpen
-            openWidgets=this.openWidgets
-            data=@data
-        }}
-    {{/each}}
-</div>
+{{#if @widgets}}
+    <div class="px-2 pb-1 d-flex justify-content-between align-items-center" ...attributes>
+        <h5 class="m-0">{{t "relatedInfo.title"}}</h5>
+        <Button
+            @type="link"
+            @action={{this.toggleAll}}
+            @label={{t (if this.showCloseAll "relatedInfo.batchClose" "relatedInfo.batchOpen")}}
+            @icon={{if this.showCloseAll "chevron-up" "chevron-down"}}
+            class="p-0 font-size-secondary"
+        />
+    </div>
+    <div class="p-2">
+        {{#each @widgets as |widget|}}
+            {{component
+                (concat "page/sidebar/widgets/" widget.widget)
+                toggleIsOpen=this.toggleIsOpen
+                openWidgets=this.openWidgets
+                data=@data
+            }}
+        {{/each}}
+    </div>
+{{/if}}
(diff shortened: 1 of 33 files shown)


pep/app/pods/components/page/sidebar/widgets/template.hbs
```diff
@@ -1,20 +1,22 @@
...
-        {{component
-            (concat "page/sidebar/widgets/" widget.widget)
-            toggleIsOpen=this.toggleIsOpen
-            openWidgets=this.openWidgets
-            data=@data
-        }}
-    {{/each}}
-</div>
\ No newline at end of file
+{{#if @widgets}}
```
bakerac4, line 1: Seems like you have two ways to hide the widgets. `hideWidgets` from above, and this. Do you need both?

pep/app/pods/components/radio-input/component.ts
```diff
@@ -0,0 +1,36 @@
...
+    inputId: string = `radio-input-${guidFor(this)}`;
+
+    /**
+     * Invoke the radio's `onChange` method if it was passed in
+     * as an argument
+     *
+     */
+    @action
+    onChange(event: Event) {
+        if (this.args.preventDefault) {
```
bakerac4, line 26: Thoughts about setting defaults here - to whatever the most common use case is

pep/app/pods/components/search/preview/template.hbs
```diff
@@ -62,7 +62,25 @@
...
                     </div>
                     <div>
-                        {{#if this.session.isAuthenticated}}
+                        {{#if @result.accessLimited}}
+                            <Alert
+                                @isShown={{this.session.isAuthenticated}}
+                                @type="warning"
+                            >
+                                {{html-safe
+                                    (or
```
bakerac4, line 71: This seems odd to me. Usually an `or` would have 3 items. Might be better as an `if`

pep/app/pods/components/page/sidebar/widgets/template.hbs
billdami: yeah, so the `{{#if @widgets}}` is for in case in the Admin configs they decide to not show any widgets on the left sidebar, in which case we dont want the "Related Info" and toggle button to show there.

the `@hideWidgets` arg is for the search route's left sidebar, where it needs to hide all the widgets if there is search results (i.e. the "Refine" section is visible).

pep/app/pods/components/radio-input/component.ts
billdami: do you mean making getters for these? like:
```ts
get preventDefault() {
    return this.args. preventDefault ?? false;
}
```
the default/most common case for both is `false`, thats why i didnt create getters, but i can if we wanna be more explicit.

for `onChange`, i can change that to just
```ts
this.args.onChange?.(this.args.value, event);
```

pep/app/pods/components/search/preview/template.hbs
billdami: not sure why an `{{or}}` would usually have 3 items? This is pretty common, its just returning the first truthy value, basically equivalent to:
```ts
return this.args.result.noAccessMessage || this.intl.t('search.preview.accessLimited')
//or
return this.args.result.noAccessMessage ?? this.intl.t('search.preview.accessLimited')
```
which we use alot

billdami:
@bakerac4 ready for another pass

PR #40, opened 2020-09-22
Story/tp66048 ip login

bakerac4:


pep/app/authenticators/ip.ts
```diff
@@ -0,0 +1,53 @@
...
+     * @param {SessionAuthenticatedData} data
+     */
+    restore(data: PepSecureAuthenticatedData) {
+        if (!isEmpty(data.SessionId)) {
+            return resolve(data);
+        } else {
+            return reject();
+        }
+    }
+}
```
billdami, line 53: @bakerac4 we probably should just keep the IP-based auth in the existing `pep` authenticator, as its more or less identical besides the `authenticate()` method, which we can customize by adding an optional `grantType: 'password' | 'ip'` arg. 

I think this is preferable, as its consistent with our patterns in other projects and how the built-in oAuth authenticator handles different "grant types", which is what this really is, just another way to auth to the same provider. But if you really want to make it a separate authenticator, we should at least extend from the `pep` authenticator, so we aren't duplicating the other logic (invalidate, restore, and also the session expiration timer logic once you add it).

Also the authenticate() method probably shouldn't have an arg for `sessionId`, since per our discussion yesterday, if we are storing that "unauthed session id" in a separate cookie, the authenticator could inject the cookies service and read it (and set it if ip-based auth "fails") as needed to keep that all self-contained.

pep/app/services/pep-session.ts
```diff
@@ -16,7 +21,53 @@ export interface AuthenticatedData {
...
+        });
+        return cookie ? JSON.parse(cookie) : undefined;
+    }
+
+    /**
+     * Clear the unauthed session data. This only happens when when the user successfully logs in
+     *
+     * @memberof PepSessionService
+     */
+    clearUnauthenticatedSession() {
```
billdami, line 65: i may have missed its usage elsewhere, but we should be using this to clear the unauthed session data cookie when either authenticators completes a successful "real" authentication.

billdami (changes requested):

pep/app/authenticators/credentials.ts
```diff
@@ -0,0 +1,164 @@
...
+import AjaxService from 'pep/services/ajax';
+import ENV from 'pep/config/environment';
+import { PepSecureAuthenticatedData } from 'pep/api';
+import { serializeQueryParams } from 'pep/utils/url';
+import PepSessionService from 'pep/services/pep-session';
+import { run } from '@ember/runloop';
+import { EmberRunTimer } from '@ember/runloop/types';
+import FastbootService from 'ember-cli-fastboot/services/fastboot';
+
+export enum SessionType {
```
billdami, line 15: thoughts on defining this in types/pep/api.d.ts? since both authenticators really "own" it equally.

pep/app/authenticators/credentials.ts
```diff
@@ -0,0 +1,164 @@
...
+     * Invalidates the local session and logs the user out
+     */
+    async invalidate(data: PepSecureAuthenticatedData) {
+        try {
+            const params = serializeQueryParams({ SessionId: data.SessionId });
+            await this.ajax.request(`${ENV.authBaseUrl}/Users/Logout?${params}`, {
+                method: 'POST',
+                headers: this.authenticationHeaders
+            });
+        } finally {
```
billdami, line 104: not 100% sure on this, but i think in order to have this resolve instead of rejecting when the request fails you need to do:
```ts
try {
    //....
} catch(err) {
   return resolve();
} finally {
    this._cancelTimeout();
}
```
as `finally` blocks don't affect/control how/what the function returns, it just gives you a spot to put logic that should always happen before it returns. You could easily test this tho by just changing the logout endpoint to an invalid url so it fails.

billdami: unless we want this to reject when the logout fails, but i thought there was a reason why we would want it to always resolve (i think due to the case where if your session needs to be invalidated after its already expired on the server, which would cause logout to fail, but we still need to successfully invalidate the local session that case.)

pep/app/authenticators/credentials.ts
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+        });
+    }
+
+    /**
+     * Schedule the session expiration to run
+     *
+     * @param {number} expiresIn
+     * @param {number} [expiresAt]
```
billdami, line 136: documented params dont looks correct here

pep/app/pods/components/page/drawer/component.ts
```diff
@@ -26,6 +27,10 @@ export default class PageDrawer extends Component<PageDrawerArgs> {
         return this.configuration.defaultSearchParams;
     }
 
+    get canLogOut() {
```
billdami, line 30: since this getter is duplicated in two places and we have our own session service now, I think it would be nice if we just defined this directly in PepSession, so you can just do `this.session.canLogOut`.

pep/app/pods/components/page/nav/component.ts
```diff
@@ -23,6 +24,10 @@ export default class PageNav extends Component<PageNavArgs> {
         return this.configuration.defaultSearchParams;
     }
 
+    get canLogOut() {
```
billdami, line 27: same here if we move this to the session service.

pep/app/services/current-user.ts
```diff
@@ -54,12 +54,10 @@ export default class CurrentUserService extends Service {
      * @returns Promise<User | void>
      */
     async fetchUser(): Promise<User | void> {
-        if (this.session.isAuthenticated) {
```
billdami, line 57: we get a User back from this endpoint even if you send in an unauthed SessionId?

And if thats the case, wouldn't you need to use the session id from `getUnauthenticatedSession()` if they dont have an authenticated session id?

billdami: related to that: I believe we need to update anywhere we are using `session.data.authenticated.SessionId` to fallback to using `getUnauthenticatedSession().SessionId` if the authenticated session ID is empty (e.g. in the ajax service, application adapter). 

Thats how the API will be able to use the session id to track the actions of unauthenticated/anonymous users.

pep/app/services/current-user.ts
```diff
@@ -54,12 +54,10 @@ export default class CurrentUserService extends Service {
...
-            const { SessionId } = this.session.data.authenticated;
-            const user = await this.store.queryRecord('user', { SessionId });
-            this.user = user;
-            return user;
-        }
+        const { SessionId } = this.session.data.authenticated;
+        const user = await this.store.queryRecord('user', { SessionId });
```
billdami, line 58: not sure if @adistasio every fully confirmed this with them. but I think we need to add the `userType` field to the user model, and then update the `updatePrefs()` method so that the api User record is NOT updated unless the `userType ===  "Individual"` (or `!== "Group"`, might be better depending on what the eventual user type will be for federated/saml logins). 

Because even if the preferences update endpoint is working for you when on an IP login, this would be bad, because all users on that IP would be sharing those same preferences -- so pref changes should be local/cookie-stored only.

billdami (changes requested):

pep/app/services/current-user.ts
bakerac4: No we don't - but this method is only requested when your logged in (in the application route)

billdami: Gotcha. Though, I think its still better to have that guard directly in the method too, because it shouldn't attempt the request no matter where its triggered from, if we wanted to call it from somewhere else in the app.

pep/app/authenticators/credentials.ts
```diff
@@ -0,0 +1,164 @@
...
+     */
+    async authenticate(username: string, password: string) {
+        try {
+            const sessionData = this.session.getUnauthenticatedSession();
+            const response = await this.ajax.request<PepSecureAuthenticatedData>(`${ENV.authBaseUrl}/Authenticate`, {
+                method: 'POST',
+                headers: this.authenticationHeaders,
+                body: this.ajax.stringifyData({
+                    grant_type: 'password',
+                    SessionId: sessionData?.SessionId,
```
billdami, line 74: @bakerac4 missed this in my last pass. but looks like `grant_type` no longer exists in the defined request body for this endpoint in the swagger docs (i think that was a holdover from the original PEP API login).

Also, it appears `SessionId` is a query param, not a body field, unless the docs are wrong on that.

pep/app/pods/components/modal-dialogs/user/login/template.hbs
```diff
@@ -41,6 +41,11 @@
                         </a>.
                     </p>
                     --}}
+                    <p class="mb-0 text-muted">
+                        <a href="https://www.psychoanalystdatabase.com/Pages/pg100Home.asp" target="_blank" rel="noopener noreferrer">
```
billdami, line 45: @bakerac4 this should be an ENV var, like `FORGOT_PW_URL`. and probably one we should leave blank in the stage/prod .env files, so that @lhorne-gavant can populate it in the deploy scripts.

pep/app/services/current-user.ts
```diff
@@ -171,7 +171,7 @@ export default class CurrentUserService extends Service {
         }
 
         // if the user is logged in, apply the new prefs locally, then save the user
-        if (this.session.isAuthenticated && this.user) {
+        if (this.session.isAuthenticated && this.user && this.user.userType !== 'Group') {
```
billdami, line 174: i think we should make this at least a unioned type, `type UserType = 'Group' | 'Individual'`  in the user model instead of comparing on the raw string directly.

pep/app/pods/components/modal-dialogs/user/login/template.hbs
bakerac4: I was under the assumption that it wouldn't change per environment? Are you sure it will?

billdami: it might not be yet, but i could easily see that happening with all the PaDS changes. i've just tried to keep all external URLs environment based so they can easily be changed if needed w/o app changes, as even if its the same across envs, it could easily change as their systems change.

If we really don't want this as an env var yet tho, i think it'd be nice to at least have it be a constant (either in the component file or a constant file).

billdami (approved):

pep/app/services/current-user.ts
adistasio: Is Individual or Group coming back?

billdami: I saw it in the swagger docs I think, but I'm not 100% sure. @bakerac4 when you get a successful IP auth response, does it actually have `UserType: "Group"` in the data?

billdami: Rather, in the GET /Users?SessionId= call is where it should be.

bakerac4: I don't have that branch running right now, but Im 90% sure it did - yes

PR #47, opened 2020-09-28
Bug/tp66328 mobile tooltip issues

billdami:


pep/app/pods/components/help-tooltip/template.hbs
```diff
@@ -0,0 +1,25 @@
+<abbr
```
bakerac4, line 1: @billdami I was working on a slightly different tooltip and I think you can clean up some of the styles associated with the search-form-control-help.

Using

```
<span class="fa-layers fa-sm">
         <FaIcon @icon="circle" @prefix="fas" class="search-results-control-help-background"/>
         <FaIcon @icon="question-circle" @size="xs" class="search-results-control-help-foreground" />
</span>
```

Which will allow you to not have to create the background yourself. It would allow you to get rid of the height, width, line-height, padding and border radius styles

billdami: @bakerac4 yea i can give that try. do you have the tooltip you were working on pushed so I can take a look? If you have something similar, ideally we can have a single base component that covers both cases, or at least covers both as closely possible w/knobs to customize as needed.

bakerac4: https://github.com/Psychoanalytic-Electronic-Publishing/PEP-Web-User-Interface/blob/7811246eed6c47c7999d52206be8202b2db7b9de/pep/app/pods/search/template.hbs#L75

bakerac4: I think the main differences for mine are that I don't want it absolutely positioned and I used the font awesome layering to create a background

billdami: thanks I'll give that a go.

Yeah, I'm keeping the absolute positioning out of the reusable `<HelpTooltip>` component (keeping that in the search form styles). So ideally you could use that same component here (since we'll want the new modal-based help text display for mobile tooltips everywhere), and then you could still apply that extra class to tweak its styles if needed.

bakerac4 (approved):

PR #56, opened 2020-09-30
Bug/tp66283 auth cookie issues - round 1

billdami:


pep/app/services/pep-session.ts
```diff
@@ -37,7 +37,6 @@ export default class PepSessionService extends SessionService.extend({
     setUnauthenticatedSession(sessionData: PepSecureAuthenticatedData) {
         const resultString = JSON.stringify(sessionData);
         this.cookies.write(UNAUTHENTICATED_SESSION_COOKIE_NAME, resultString, {
-            domain: ENV.cookieDomain,
             secure: this.isCookieSecure,
             sameSite: ENV.cookieSameSite,
             expires: new Date('2525-01-01') // never!!!
```
bakerac4, line 42: I think it would be beneficial to bring this, and the secure values over into the cookie service since they are used in multiple places. So it would end up looking like

```
this.cookies.write(UNAUTHENTICATED_SESSION_COOKIE_NAME, resultString, {
    secure: this.cookies.isCookieSecure,
    expires: this.cookie.expirationDate
...
```

pep/node/server.js
```diff
@@ -18,7 +18,7 @@ module.exports = function(emberDistPath) {
         fastbootMiddleware({
             distPath: emberDistPath,
             //resilient mode = true swallows rendering errors and returns a 200 w/the default index.html
-            resilient: false
+            resilient: true
```
bakerac4, line 21: What are we trying to fix with this? Initially this looks like its just going to hide the actual errors from us when we would want to fix the root cause?

billdami: yes and no. this will (theoretically) be a safety net for errors/failed api requests that we don't account for, and the user will get served the traditional non-fastboot rendered index page, and when the app renders client-side, it would produce the same errors, but within the app UI. This is mainly to protect against the hard crashed response state where the user just sees a blank white page with the `{ error: "Internal server error"}` json that apigateway spits back normally if Fastboot fails to render the page. We never want the user to get that in prod.

Also note that this change only applies when running the prod fastboot server, so during dev if there's fastboot related errors they wouldn't be hidden and we should still make sure we are handling them. I'm still looking into the specific error thats causing the white error pages for us now, but wanted to try adding this as well as an extra layer of protection so the app more gracefully degrades to a traditional non-server rendered SPA in cases were fastboot just bombs out completely (but yes i agree these errors should still be tracked and handled at their source when possible).

pep/app/services/pep-session.ts
billdami: we dont have our own `cookies` service defined, its just using the addon's service directly.  i guess i could create one, it would need to be named differently due to the .js/.ts collision issue, but id rather save that for a clean up pass in my next branch so as not to hold these fixes up.

bakerac4: Maybe a half fix is just moving the expires into a constant and creating a global helper for isCookieSecure

billdami: yeah i can do that. actually, `cookieSecure` should just really be converted to boolean directly when its set in environment.js:
```js
cookieSecure: Number(process.env.COOKIE_SECURE) === 1,
```

billdami:
@bakerac4 updated.

PR #70, opened 2020-10-13
Fixing forwarded for issues

bakerac4:


pep/app/utils/url.ts
```diff
@@ -101,3 +100,20 @@ export function documentCSVUrl(store: DS.Store, queryParams: QueryParamsObj) {
...
+ *
+ * @export
+ * @param {*} headers
+ * @returns {string}
+ */
+export function parseForwardedForHeaders(headers?: any) {
+    if (headers) {
+        const xForwardedFor: string[] = headers.get('X-Forwarded-For') ?? [''];
+        const ipAddress = xForwardedFor[0].split(',');
+        return ipAddress[0];
```
billdami, line 115: i dont think we should be splitting the string to only send the first ip, since thats changing the format, sending the full list is what the standard is. we should just be passing along the true full value, and then pads can decide how to parse it.

> `<client>`
> The client IP address
> 
> `<proxy1>, <proxy2>`
> If a request goes through multiple proxies, the IP addresses of each successive proxy is listed. This means, the right-most IP address is the IP address of the most recent proxy and the left-most IP address is the IP address of the originating client.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

billdami (changes requested):

billdami (approved):

PR #81, opened 2020-10-16
Fixing search issues

bakerac4:


pep/app/pods/components/search/form/term-control/component.ts
```diff
@@ -26,4 +27,8 @@ export default class SearchFormTermControl extends Component<SearchFormTermContr
     get termContentConfig() {
         return this.args.searchTerm ? this.configuration.content.search.terms.types[this.args.searchTerm.type] : null;
     }
+
+    get canRemove() {
+        return typeof this.args.canRemove === 'boolean' ? this.args.canRemove : true;
```
billdami, line 32: any reason this can't be:
```ts
return this.args.canRemove ?? true;
```

billdami (changes requested):

billdami (approved):

PR #83, opened 2020-10-20
Feedback 2 + starting doc formatting

bakerac4:


pep/app/pods/document/model.ts
```diff
@@ -1,11 +1,12 @@
...
 import DS from 'ember-data';
 import attr from 'ember-data/attr';
 import { belongsTo } from 'ember-data/relationships';
 
-import { isEmpty } from '@ember/utils';
-
-import { INVALID_ABSTRACT_TAGS, INVALID_ABSTRACT_PREVIEW_TAGS, HTML_BODY_REGEX } from 'pep/constants/regex';
+import { HTML_BODY_REGEX, INVALID_ABSTRACT_PREVIEW_TAGS, INVALID_ABSTRACT_TAGS } from 'pep/constants/regex';
 import SimilarityMatch from 'pep/pods/similarity-match/model';
+import { parseXML } from 'pep/utils/dom';
```
billdami, line 9: doesn't look like this is being used in the file.

billdami (changes requested):

pep/app/pods/components/document/text/component.ts
```diff
@@ -18,9 +20,59 @@ export default class DocumentText extends Component<DocumentTextArgs> {
     @service loadingBar!: LoadingBarService;
     @service modal!: ModalService;
 
+    @tracked xml?: XMLDocument;
+
+    constructor(owner, args) {
+        super(owner, args);
+        const XML = parseXML(this.args.text);
```
billdami, line 27: we should guard against running any of this in fastboot at all, so that it doesnt even generate an `Error`. maybe move all this logic to a method with the `@dontRunInFastboot` decorator, and call that from the constructor.

PR #88, opened 2020-10-21
Story/tp65877 reading pane

bakerac4:


pep/app/constants/themes.ts
```diff
@@ -7,18 +7,54 @@ export interface Theme {
...
+        links: {
+            glossary: string;
+            bibliography: string;
+            pageReference: string;
+            external: string;
+            figure: string;
+            webExternal: string;
+            webInternal: string;
+            special: string;
+        };
```
billdami, line 20: unless we have a good reason to do these in-JS, i think we should try to keep these defined directly in the SASS theme files so its part of the normal compiled CSS, like the rest of the theming is done currently. that gives us better integration w/existing bootstrap styles and reduces complexity/additional deps.

if the concern was making these values editable via admin configs, I had some discussions w/Alan about that early on, and would make the case that manual theme editing/control like that at a granular per element/color level might not be the best path (unintended design problems, accessibility issues, etc), as opposed to just having a the themes library managed and adjusted by devs in the codebase itself.

pep/app/pods/components/document/text/component.ts
```diff
@@ -32,52 +35,34 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+        if (!(xml instanceof Error)) {
+            const xslt = await this.loadXSLT();
 
-            this.convertBibliography(XML);
-            this.convertPageBreaks(XML);
-            this.xml = XML;
+            if (xslt && document.implementation && document.implementation.createDocument) {
+                const processor = new XSLTProcessor();
+                const colors = this.theme.currentTheme.colors.links;
+                processor.setParameter('', 'glossaryColor', colors.glossary);
```
billdami, line 55: this looks like the one spot where those colors are actually needed in-JS,  I don't see where that `glossaryColor` param is actually used in the XSLT file, and we are just setting colors with generated CSS/classes anyway, unless I'm missing something?

billdami (changes requested):

pep/app/constants/themes.ts
bakerac4: Yes, these values per the spec are suppose to be editable from the admin. We can always adjust accordingly however for now, I think this should stay as is

PR #92, opened 2020-10-23
Story/tp65877 reading pane

bakerac4:


pep/app/pods/components/document/text/component.ts
```diff
@@ -15,6 +16,7 @@ import GlossaryTerm from 'pep/pods/glossary-term/model';
 import LoadingBarService from 'pep/services/loading-bar';
 import ThemeService from 'pep/services/theme';
 import { parseXML } from 'pep/utils/dom';
+import tippy from 'tippy.js';
```
billdami, line 19: we couldn't reuse popper.js for this? :sadpanda:

pep/package.json
```diff
@@ -48,6 +48,7 @@
         "@glimmer/component": "^1.0.1",
         "@glimmer/tracking": "^1.0.0",
         "@html-next/vertical-collection": "^1.0.0",
+        "@popperjs/core": "^2.5.3",
```
billdami, line 51: you still have the dep added here even though it looks like its not being used

billdami (changes requested):

PR #101, opened 2020-10-27
Adding read link functionality

bakerac4:


pep/app/pods/components/page/nav/template.hbs
```diff
@@ -32,7 +32,7 @@
         {{! TODO project reqs state that there should be a "Read" nav item         }}
         {{! but where does it go/show? Seems like it would be redundant w/"Browse" }}
```
billdami, line 33: remove outdated comments

pep/app/pods/components/page/nav/template.hbs
```diff
@@ -32,7 +32,7 @@
...
         <li class="nav-item">
-            <LinkTo @route="read" class="nav-link">
+            <LinkTo @route="read" @disabled={{this.readDisabled}} class="nav-link">
```
billdami, line 35: maybe not doable for this initial merge, but I think we should at least consider having this link go to some sort of landing/"empty state" page (e.g. w/some content that states "Your last viewed document will be accessible here...", similar to the old "empty" search page layout maybe), instead of being outright disabled.

This is just a highly unusual pattern from a pure UX perspective. As a new user coming into the site fresh, I see a top-level "Read" link front and center in the main navigation, but its disabled...why? The user has no context or idea what that link should do.

cc @adistasio

pep/app/pods/read/index/route.ts
```diff
@@ -16,4 +20,10 @@ export default class ReadIndex extends Route {
 
         copySearchToController(appController);
     }
+
+    redirect(model: any, transition: Transition) {
+        if (transition.to.name === this.routeName && this.currentUser.lastViewedDocumentId) {
```
billdami, line 25: verify this, but im 99% sure you don't need this `transition.to.name === this.routeName` check here, as the `redirect()` will only fire if the user's final destination route is indeed read/index (i.e. it shouldn't fire if you are transitioning to read/document.

pep/app/pods/application/controller.ts
```diff
@@ -43,6 +43,7 @@ export default class Application extends Controller {
     @tracked viewedCount: string = '';
     @tracked viewedPeriod: ViewPeriod = ViewPeriod.PAST_WEEK;
     @tracked searchTerms: SearchTermValue[] = [];
+    @tracked lastViewedDocument?: Document;
```
billdami, line 46: doesnt look like this is being used

billdami (changes requested):

pep/app/pods/components/page/drawer/component.ts
```diff
@@ -19,13 +21,18 @@ export default class PageDrawer extends Component<PageDrawerArgs> {
...
+    @service currentUser!: CurrentUserService;
 
     @tracked isUserMenuOpen = false;
 
     get defaultSearchParams() {
         return this.configuration.defaultSearchParams;
     }
 
+    get readDisabled() {
+        return !!!this.currentUser.lastViewedDocumentId;
```
billdami, line 33: @bakerac4 missed this, but we should fix this on the next branch since its unnecessary and the page/nav/component's version correctly only does `return !this.currentUser.lastViewedDocumentId;`

PR #103, opened 2020-10-28
Fixing terms and phrases saving

bakerac4:


pep/app/utils/search.ts
```diff
@@ -262,13 +262,18 @@ export function groupCountsByRange(
  */
 export function copySearchToController(toController: Controller & SearchController) {
     const searchController = getOwner(toController).lookup(`controller:search`);
+    const config: ConfigurationService = getOwner(toController).lookup('service:configuration');
+    const defaultFields = config.base.search.terms.defaultFields;
```
billdami, line 266: will this pull in the user's previously saved search term field preferences to use instead of the admin config defaults, when loading/reloading the app on the homepage directly?

bakerac4: Yes - it gets it from the search controller (which does that part)

billdami: I'm only seeing that happen if you load the app on the /search route directly on your branch, if i load the app directly on the homepage, it just renders the admin config's default fields. Which I think is because the search controller only gets the user prefs' term fields when its `setupController()` is run, which only happens if you actually enter that route.

pep/app/authenticators/credentials.ts
```diff
@@ -103,6 +105,7 @@ export default class CredentialsAuthenticator extends BaseAuthenticator {
     async invalidate(data: PepSecureAuthenticatedData) {
         try {
             const params = serializeQueryParams({ SessionId: data.SessionId });
+            this.currentUser.clearPreferences();
```
billdami, line 108: why are we clearing preferences on logout, was this in the spec?

based on my past discussions w/ @adistasio, i was under the impression that local preferences (cookies, local storage) should always be preserved, regardless of if the user is logged in or out.

bakerac4: Not sure, but I think it makes the most sense. If you log out, we shouldn't be showing your theme choice and other choices you made - we should be showing the default one.

billdami: yeah so we discussed that a bit, but we came to the conclusion that prefs (local-stored ones at least) should just always stick around. One reason was that you can set preferences like theme and search fields while logged out, and they will be preserved. So would seem odd that if you then happened to log in at some point later, and then logged out, that those prefs would be reset, since they never got reset before as long as you stayed logged out.

bakerac4: I still think it might be better to give the users a clean slate. If you share a computer with a colleague and he logs out, you don't want to have to go and change all his settings. That way you have a consistent reproducible logged out state each time @adistasio

adistasio: Let's leave it as is, don't clear preferences on logout.

PR #107, opened 2020-11-04
Story/tp65877 center pane

bakerac4:


pep/ember-cli-build.js
```diff
@@ -5,7 +5,8 @@ const EmberApp = require('ember-cli/lib/broccoli/ember-app');
 module.exports = function(defaults) {
     let app = new EmberApp(defaults, {
         fingerprint: {
-            prepend: process.env.ASSETS_BASE_URL
+            prepend: process.env.ASSETS_BASE_URL,
+            extensions: ['js', 'css', 'png', 'jpg', 'gif', 'map', 'xslt']
```
billdami, line 9: did you test if adding the extension here alone will fingerprint the file in a prod build? I thought that ember doesn't auto-fingerprint dynamically generated paths that are in JS-code, only paths in templates, and you needed to use a specific ember-cli api to get the fingerprinted url in those cases. I could be wrong on that tho.

pep/app/pods/read/document/controller.ts
```diff
@@ -354,6 +356,29 @@ export default class ReadDocument extends Controller {
...
+     *
+     * @param {string} searchTerms
+     * @memberof ReadDocument
+     */
+    @action
+    viewSearch(searchTerms: string) {
+        // TODO improve this typing
+        clearSearch(this as any, this.configuration, this.currentUser);
+        this.router.transitionTo('search', {
+            queryParams: {
```
billdami, line 371: you could use the configuration service's `defaultSearchParams` getter as the basis for this to dry it up a little:
```ts
queryParams: { ...this.configuration.defaultSearchParams, searchTerms }
```

pep/app/pods/components/page/sidebar/widgets/whats-new/component.ts
```diff
@@ -49,7 +50,9 @@ export default class PageSidebarWidgetsWhatsNew extends Component<PageSidebarWid
     @action
     @dontRunInFastboot
     onElementInsert() {
-        taskFor(this.loadResults).perform();
+        setTimeout(() => {
+            taskFor(this.loadResults).perform();
+        }, 1000);
```
billdami, line 55: i know this is probably only temporary, but we should use `later(() => ..., 1000);`

pep/app/pods/components/page/sidebar/widgets/most-viewed/component.ts
```diff
@@ -49,7 +51,9 @@ export default class PageSidebarWidgetsMostViewed extends Component<PageSidebarW
     @action
     @dontRunInFastboot
     onElementInsert() {
-        taskFor(this.loadResults).perform();
+        setTimeout(() => {
+            taskFor(this.loadResults).perform();
+        }, 3000);
```
billdami, line 56: samesies, use `later()`

pep/app/pods/components/page/sidebar/widgets/most-cited/component.ts
```diff
@@ -49,7 +50,9 @@ export default class PageSidebarWidgetsMostCited extends Component<PageSidebarWi
     @action
     @dontRunInFastboot
     onElementInsert() {
-        taskFor(this.loadResults).perform();
+        setTimeout(() => {
+            taskFor(this.loadResults).perform();
+        }, 2000);
```
billdami, line 55: samesies, use `later()`

pep/app/pods/application/route.ts
```diff
@@ -63,12 +64,23 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
...
+                try {
+                    await this.session.authenticate('authenticator:ip');
+                } catch (errors) {
+                    this.session.invalidate();
+                }
+                // If you are logged in, but your currently authenticated using IP - try again to ensure you didn't change locations
+                // But here we don't wait for the promise to return as that would cause a FOUC
+            } else if (this.session?.data?.authenticated?.SessionType === SessionType.IP) {
+                try {
+                    this.session.authenticate('authenticator:ip');
```
billdami, line 80: how would `await`ing this cause a FOUC if its only being run in fastboot?

pep/app/pods/components/document/text/component.ts
```diff
@@ -31,22 +67,47 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+        onClickOutside(instance: Instance<Props>) {
+            instance.hide();
+        }
+    };
+
     constructor(owner: unknown, args: DocumentTextArgs) {
         super(owner, args);
-        this.parseDocumentText(args.text);
+        this.parseDocumentText(
+            args.document.accessLimited ? args.document.abstractCleaned : args.document.documentCleaned
```
billdami, line 88: could we ditch these `*Cleaned` getters entirely now, since we are using the XML versions?

pep/app/pods/components/document/text/component.ts
```diff
@@ -76,23 +144,97 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+                url,
+                caption,
+                id: parseInt(figureId?.substring(1) ?? '', 10)
+            });
+        } else if (type === DocumentLinkTypes.AUTHOR_SEARCH) {
+            const firstName = target?.querySelector('.nfirst')?.innerHTML;
+            const lastName = target?.querySelector('.nlast')?.innerHTML;
+            const name = `${firstName} ${lastName}`;
+            this.args.viewSearch(JSON.stringify([{ term: name, type: SearchTermId.AUTHOR, value: name }]));
         }
```
billdami, line 224: maybe not for this merge, but would be nice if we could break up this mammoth method, maybe with helper methods/functions for handing each of the different link types, and use a hash to map each type to its respective method, instead of a ginormous if/else block?

pep/app/pods/components/document/text/component.ts
```diff
@@ -103,40 +245,96 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+     *
+     * @param {HTMLElement} element
+     * @memberof DocumentText
+     */
     @action
     setupListeners(element: HTMLElement) {
         this.containerElement = element;
         scheduleOnce('afterRender', this, this.attachTooltips);
+        if (this.args.page) {
+            this.scrollToPage(parseInt(this.args.page, 10));
```
billdami, line 284: use `Number(val)` instead of `parseInt(val, 10)` unless `val` is known to contain some garbage/non-number characters that you need to filter out. `parseInt()` handles that with more complex parsing, but in most cases you only need the simple casting/coercing that `Number()` provides.

billdami (changes requested):

pep/app/pods/application/route.ts
bakerac4: Im not 100% sure, but it does. I believe the FOUC actually happens in FastBoot.

pep/app/pods/components/document/text/component.ts
bakerac4: Yea, there is garbage in the values that needs to be converted.

pep/app/pods/components/document/text/component.ts
bakerac4: Possibly, I'll make that change in 005 which will give me time to play with it and make sure nothing breaks

pep/app/pods/components/document/text/component.ts
bakerac4: I'll look into this when I have time. I've added a TODO

pep/app/pods/application/route.ts
billdami: does the FOUC not happen in the above case though, where if you are not logged in, we do `await` the authenticate call? I had seen something similar on stage recently though that sounds similar, where the site will appear to render w/o the stylesheet loaded, not sure if its related, but maybe.

Also, if we can't `await` here, we should remove the try/catch around it, as since you are not awaiting the response, the `catch` will never be hit if it fails. 

(also this seems like it could result in a race condition possibly, since its not blocking the fastboot request from returning, the response could be sent to the client before this has a chance to set cookies/etc, seems a bit weird.)

billdami (changes requested):
just the 1 follow up comment on `application/route.ts`

pep/app/pods/application/route.ts
bakerac4: Correct. So if you are not logged in and try the IP auth there is no FOUC, if you are logged in and try the IP auth it does. Which is why I put it as a separate block

billdami: yea that part is fine for now (though I think we should figure out the real issue of why FOUC happens in one case but not the other when we get some time).

but i think we should at least remove the try/catch on the one that doesnt have the `await`, because its not doing anything. since you are not waiting for a response, a failed authenticate request will never hit that `catch`.

bakerac4: That second else if is checking if you are still logged in correctly. So if your logged into IP already, we try to re auth you. In 99% of cases, there will be no change and in the 1% that there is, we should just log you out when that comes back

PR #117, opened 2020-11-12
Story/tp65878 browse selection pane

bakerac4:


pep/app/utils/dom.ts
```diff
@@ -79,3 +79,17 @@ export function parseXML(input: string): XMLDocument | Error {
...
+ *
+ * @return {XSLT}
+ */
+export function loadXSLT(this: any): Promise<Document | null> {
+    return new Promise((resolve) => {
+        let request = new XMLHttpRequest();
+        request.open('GET', `/xmlToHtml.xslt`, false);
+        request.send('');
+        resolve(request.responseXML);
+    });
```
billdami, line 94: I meant to ask this way back when it was first implemented, and more curiosity then anything, but how come we went with a native browser XMLHttpRequest (and a non-async, blocking request at that), over just a normal fetch(), which should be able to handle XML responses too?

Also I think this will be a blocker if we ever want to get document rendering working in fastboot, since there is no XMLHttpRequest in node.

billdami (changes requested):

pep/app/pods/browse/journal/route.ts
```diff
@@ -0,0 +1,7 @@
+import Route from '@ember/routing/route';
+
+export default class BrowseJournal extends Route.extend({
+  // anything which *must* be merged to prototype here
+}) {
```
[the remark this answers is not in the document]
bakerac4: All of these are for future pages (just stubs) - but I can delete the classic extend

pep/app/utils/dom.ts
bakerac4: When I tried it, It didn't work. However I was able to get it to work this time. 

That being said, we wont be able to render in fastboot anyways without changes because `fetch` does not natively handle XML, but we can convert to text and then use a parser to parse the string.

pep/app/utils/dom.ts
```diff
@@ -79,3 +79,18 @@ export function parseXML(input: string): XMLDocument | Error {
...
+ */
+export function loadXSLT(this: any): Promise<Document | null> {
+    return new Promise((resolve) => {
+        fetch('/xmlToHtml.xslt').then(async (response) => {
+            const text = await response.text();
+            const xml = parseXML(text);
+            resolve(xml as Document);
+        });
+    });
+}
```
billdami, line 96: I think you'll want to do `import fetch from 'fetch';` in this file so it uses the version of `fetch()` coming from ember-fetch which is isomorphic and works in node as well.

also, don't have to change this if you don't want to, but I think this whole function could be async/await based couldn't it? e.g.

```ts
export async function loadXSLT(this: any): Promise<Document | null> {
    const response = await fetch('/xmlToHtml.xslt');
    const text = await response.text();
    const xml = parseXML(text);
    return xml as Document;
}
```

billdami (changes requested):

pep/app/utils/dom.ts
billdami: yeah good point, from what I'm seeing, to have full fastboot compatibility it looks like we'd have to treat it as text and use a third party xml parsing lib that is node-friendly.

billdami (approved):

PR #123, opened 2020-11-19
Story/tp65879 browse center pane

bakerac4:


pep/app/pods/browse/journal/volume/route.ts
```diff
@@ -0,0 +1,13 @@
+import Route from '@ember/routing/route';
+
+export default class BrowseJournalVolume extends Route {
+    model(params: { volume_number: string }) {
+        const journalParams = this.paramsFor('browse.journal') as { pep_code: string };
```
billdami, line 5: just cleanup, but in other projects we usually type the params as an exported interface on the route file that it belongs to e.g.
```ts
// in browse/journal/route.ts
export interface BrowseJouralParams {
    pep_code: string
}
```

pep/app/styles/pages/browse.scss
```diff
@@ -0,0 +1,21 @@
...
+    }
+    .dotted-line {
+        position: absolute;
+        border-bottom: dotted 1px $black;
+        width: 100%;
+        height: 100%;
+    }
+
+    .page {
+        width: 35px;
```
billdami, line 14: should probably use `rem` for dimensions like this when we can (i.e. this is probably about `2rem`). when `rem` is used universally for styles it makes scaling the UI to different base font sizes (which we'll eventually need for the user font size preference) much easier.

pep/app/pods/browse/journal/index/route.ts
```diff
@@ -0,0 +1,44 @@
...
+            context: controller,
+            modelName: 'volume',
+            filterList: ['sourcecode'],
+            models: model.toArray() ?? [],
+            metadata: model.meta,
+            pagingRootKey: null,
+            filterRootKey: null
+        });
+    }
+}
```
billdami, line 44: missing doc blocks in this file

pep/app/pods/browse/journal/index/route.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import BrowseJournalIndexController from 'pep/pods/browse/journal/index/controller';
+import Journal from 'pep/pods/journal/model';
+import Source from 'pep/pods/source/model';
+
+export default class BrowseJournalIndex extends Route {
+    sourceCode?: string;
+    journalInformation?: Journal;
+    model() {
+        const journalParams = this.paramsFor('browse.journal') as { pep_code: string };
```
billdami, line 14: continuing from the other comment, you'd then be able to reuse that params interface and import it here

pep/app/pods/components/modal-dialogs/help-tooltip/template.hbs
```diff
@@ -3,15 +3,15 @@
     @size="sm"
     @overlayClassNames="slide-card-overlay"
     @modal-class="slide-card-modal"
-    as |Modal|
+    @closable={{false}} as |Modal|
```
billdami, line 6: fyi this modal is only used in mobile and it has a very different layout (its a "slide up card" type UI), so we might want to keep closable=true here for its improved UX, unless they explicitly asked to remove close-on-outside-tap behavior here too.

pep/app/pods/read/document/controller.ts
```diff
@@ -133,6 +142,26 @@ export default class ReadDocument extends Controller {
...
+     * @memberof Search
+     */
+    @action
+    async onChangeSorting(sorts: string[]) {
+        if (sorts.length) {
+            return transformSearchSortToAPI(sorts);
+        } else {
+            return [];
+        }
+    }
```
billdami, line 159: unnecessary `async`

could we move the 
```ts
if(!sorts.length) {
    return [];
}
```
logic into the `transformSearchSortToAPI()` util function, turning this into just
```ts
@action
onChangeSorting(sorts: string[]) {
    return transformSearchSortToAPI(sorts);
}
```

billdami (changes requested):

pep/app/pods/components/modal-dialogs/help-tooltip/template.hbs
bakerac4: They stated do it globally, so Im going to leave this for now and if they want it changed back we can easily do that

pep/app/pods/read/document/controller.ts
bakerac4: IMO I don't think we should be calling that function if we don't have any sorts. What about

```
return sorts.length ? transformSearchSortToAPI(sorts) : [];
```

billdami: if you are doing this same guard in multiple places where thats used i think its worthwhile to have the logic inside it, to just make it more robust it how it handles the args its given, its not like you are breaking its type contract, you still are passing it `string[]`.

that being said, i'm fine with your proposed change above.

billdami (approved):

PR #127, opened 2020-11-24
Change print format depending on display selection

bakerac4:


bakerac4:
Fixes #127

pep/app/pods/read/document/controller.ts
```diff
@@ -364,24 +364,29 @@ export default class ReadDocument extends Controller {
...
-            }
-        ]);
+        if (this.selectedView.id === SearchViewType.BIBLIOGRAPHIC) {
+            const html = this.printer.dataToBibliographicHTML(data);
+            this.printer.printHTML(html);
+        } else {
+            this.printer.printJSON<Document>(data, [
+                {
+                    field: 'authorMast',
+                    displayName: 'Author'
```
billdami, line 374: these display names should probably be translated

pep/app/services/printer.ts
```diff
@@ -1,8 +1,23 @@
...
+     * @param {Document[]} data
+     * @return {*}
+     * @memberof PrinterService
+     */
+    dataToBibliographicHTML(data: Document[]) {
+        const listItemsHtml = data.map((item) => {
+            return `<li class="mb-3 print-item">${item.authorMast}, ${item.year}, ${item.title}, ${item.documentRef}</li>`;
+        });
+        const html = `<div><ul class="list-unstyled">${listItemsHtml.join('')}</ul></div>`;
+        return html;
```
billdami, line 19: probably not worth it yet, but would be nice if we could utilize the glimmer primitives/API, and be able to put this html in a .hbs template, and then load it and render it here w/glimmer.

pep/app/services/printer.ts
```diff
@@ -11,13 +26,27 @@ export default class PrinterService extends Service {
...
+     * Print data using [PrintJS](https://printjs.crabbly.com/)
+     *
+     * @param {string} data
+     * @memberof PrinterService
+     */
+    printHTML(data: string) {
+        printJS({
+            printable: data,
+            type: 'raw-html',
+            style: `.list-unstyled { padding-left: 0; list-style: none; font-size: 12px; } .mb-3 { margin-bottom: 1rem !important; }`
```
billdami, line 47: maybe not worth it yet, but if we ever have to add/duplicate more styles to this, we should consider just doing a fetch() of the `/vendor.scss` as text and throw it in here, so you have all the bootstrap styles at your disposable. and maybe even cache the styles in this service so it only needs to be loaded once per session.

billdami (changes requested):

pep/app/services/printer.ts
bakerac4: I actually initially just imported the pep and vendor.css but then the paging didn't work. I didn't want to spend hours trying to figure out what broke it so I went this way and just included what I needed

billdami: whats the paging? But yeah loading the `pep.css` might've been the culprit (or part of it) since it has a bunch of global layout styles and stuff that probably would muck with print display majorly.

Though i just realized that we import Bootstrap in our app (pep.css) file, so just using the vendor.css only wouldn't work anyway.. we'd need to think of a way to pull that apart, or maybe just add some global print styles that override the stuff that is causing it to break in print.

billdami (approved):

Issue #63, opened 2020-10-06
Word Wheel


nrshapiro:
The word wheel works nicely if you type slowly enough.  However, when you know what you want to enter, and just type it at normal speed and press enter at the end, the wordwheel often pops up afterward and even stays up (most of the time?), so you have to go click to dismiss it.

Once the user presses Enter (or the search button if it's ever that slow), the word wheel event should be cancelled, and if already up, the word wheel itself dismissed.

adistasio:
I believe this to be resolved.

nrshapiro:
![2020-12-17_14h29_44](https://user-images.githubusercontent.com/4837554/102534574-fc682600-4074-11eb-8eda-b558e74f78f2.gif)

bakerac4:
@nrshapiro I believe this is been fixed somewhere along the way. Can you confirm?

nrshapiro:
Retested...popdown box now closes when you press enter.  That works.

PR #129, opened 2020-11-25
Fixing author word wheel

bakerac4:


bakerac4:
Fixes #63

pep/app/pods/components/word-wheel/component.ts
```diff
@@ -34,16 +39,20 @@ export default class WordWheel extends Component<WordWheelArgs> {
...
      * @param {string} currentWord
      * @returns {Promise<FlTypeaheadSuggestion[]>}
      */
     @action
-    async loadSuggestions(currentWord?: string): Promise<FlTypeaheadSuggestion[]> {
+    async loadSuggestions(currentWord?: string, text?: string): Promise<FlTypeaheadSuggestion[]> {
         if (currentWord?.length) {
             const params = removeEmptyQueryParams({
-                word: currentWord,
+                word: this.searchType === WordWheelSearchType.WORD ? currentWord : text,
```
billdami, line 55: I probably would've done this at the core `fl-typeahead` component level that the word-wheel is built on, as it would allow it to completely skip a lot of the complex logic it is doing to provide and insert that `currentWord` as the user types, but this is fine.

PR #130, opened 2020-11-25
Login issue

### LINKED ISSUE #109: Login Issue - Need more info
Today's problems with PaDS Login point to an issue.  When I try to login, with a valid account, it just tells me it can't login.  Really, the user needs to know to stop trying, that login's aren't available at that time, to try again later.

Not the best news for the user, but not sure what else we can do at this point.  But what we don't want is the user to be trying other things, calling support, etc., to know why his login isn't working...when in fact, we already know that it's a problem elsewhere.

bakerac4:
Fixes #109

### DIFF
--- a/pep/translations/en-us.json
+++ b/pep/translations/en-us.json
@@ -485,7 +485,7 @@
         "clickToSubscribe": "Click here to subscribe",
         "success": "You are now logged in.",
         "forgotPassword": "Forgot Password?",
-        "genericError": "Unable to login. There is a problem behind the scenes and we are looking into it. Please try again later"
+        "genericError": "Unable to login. There is a problem behind the scenes, and we are looking into it. Please try again later."
     },
     "preferences": {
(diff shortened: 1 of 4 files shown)


billdami:
@bakerac4 is this PR just named incorrect?

pep/translations/en-us.json
```diff
@@ -484,7 +484,8 @@
         "notAMember": "Not a member?",
         "clickToSubscribe": "Click here to subscribe",
         "success": "You are now logged in.",
-        "forgotPassword": "Forgot Password?"
+        "forgotPassword": "Forgot Password?",
+        "genericError": "Unable to login. There is a problem behind the scenes and we are looking into it. Please try again later"
```
billdami, line 488: comma after `scenes`, and missing period on last sentence.

billdami (changes requested):

billdami (approved):

PR #137, opened 2020-11-30
Fixing hic limit saving

### LINKED ISSUE #99: Hits in Context Limit set in my preferences is reset when I refresh
I am logged in
I set the hits in context limit to 5.
I do a search, it works fine.
I do a browser refresh (F5), it's set to 1 again.
shouldn't that be set to my user settings?  (Which change after the refresh!)
I think it's ok for non-logged in users just to set it in cookies.  Serves them right for not logging in ;-)  (Though cookies aren't reset by F5, so I'm not sure why it resets in any case)

bakerac4:
Fixes #99

### DIFF
--- a/pep/app/pods/components/modal-dialogs/user/preferences/template.hbs
+++ b/pep/app/pods/components/modal-dialogs/user/preferences/template.hbs
@@ -33,6 +33,7 @@
                         @placeholder={{t "preferences.search.hic.limit"}}
                         @value={{this.currentUser.preferences.searchHICLimit}}
                         maxlength="2"
+                        {{on "input" (fn this.updatePreference this.hicLimit)}}
                     />
                 </div>
             </div>
--- a/pep/app/pods/components/modal-dialogs/user/preferences/component.ts
+++ b/pep/app/pods/components/modal-dialogs/user/preferences/component.ts
@@ -2,6 +2,9 @@ import { action } from '@ember/object';
 import { inject as service } from '@ember/service';
 import Component from '@glimmer/component';
 
+import { timeout } from 'ember-concurrency';
+import { restartableTask } from 'ember-concurrency-decorators';
+import { taskFor } from 'ember-concurrency-ts';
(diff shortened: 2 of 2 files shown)


pep/app/pods/components/modal-dialogs/user/preferences/template.hbs
```diff
@@ -33,6 +33,7 @@
                         @placeholder={{t "preferences.search.hic.limit"}}
                         @value={{this.currentUser.preferences.searchHICLimit}}
                         maxlength="2"
+                        {{on "input" (fn this.updatePreference this.hicLimit)}}
```
billdami, line 36: since these updates are hitting the api if you are logged in we should at least debounce these updates from free text fields with a concurrency task I think. Otherwise switch to a changeset/Save button flow for saving updates rather than updating as input values change.

billdami (changes requested):

pep/app/pods/components/modal-dialogs/user/preferences/template.hbs
bakerac4: Not sure if thats really necessary since the field has a `maxlength` of 2. I could see that being a nice to have down the line if we have extra time?

billdami: thats true, its not a _huge_ deal here (though you'd still get excess requests if the user types in like "10", then backspaces to "1" -- 3 api requests instead of 1 if debounced), but i'd just hate to get in that habit or moreso have patterns in place that other devs who come onto the project might copy in places where it definitely is more necessary.

Plus I mean its very low hanging fruit, would just add a couple more lines to the component if we did something like:
```ts
@restartableTask
*debouncedUpdatePreference(/*...*/) {
    yield timeout(250);
    yield this.updatePreference(/*...*/);
}
```
```hbs
{{on "input" (perform this.debouncedUpdatePreference this.hicLimit)}}
```

billdami (approved):

PR #139, opened 2020-11-30
Adding session ID to images

### LINKED ISSUE #93: Graphics Callouts with Session-id (client-session)
I was thinking about the embedded graphics callouts, and remembered you need to be injecting (replacing) the url in those so that they work...and I realized that, at the same time, you could inject the client-id and client-session into the parameters.  That would be helpful if James needs to track graphics in counter reports, would allow me to restrict graphics to logged in users, as it was before, and would clean up the logging a bit.

Example callout:

`Journal Logo`

with a session id would be:
`Journal Logo`

Perhaps not a great example, since it's just a banner, but it's the same endpoint call for the image.

bakerac4:
Fixes #93

### DIFF
--- a/pep/tests/integration/pods/components/image/component-test.ts
+++ b/pep/tests/integration/pods/components/image/component-test.ts
@@ -1,26 +1,28 @@
-import { module, test } from 'qunit';
-import { setupRenderingTest } from 'ember-qunit';
 import { render } from '@ember/test-helpers';
+
+import { setupRenderingTest } from 'ember-qunit';
+
 import hbs from 'htmlbars-inline-precompile';
+import { module, test } from 'qunit';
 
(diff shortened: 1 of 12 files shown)


pep/tests/integration/pods/components/image/component-test.ts
```diff
@@ -0,0 +1,28 @@
...
+        await render(hbs`{{image}}`);
+
+        assert.equal(this.element.textContent?.trim(), '');
+
+        // Template block usage:
+        await render(hbs`
+      {{#image}}
+        template block text
+      {{/image}}
+    `);
```
billdami, line 24: this block form test is prob failing since the component doesnt `{{yield}}`

billdami (changes requested):

billdami (approved):

Issue #72, opened 2020-10-14
Need to display the full-text in the middle pane with navigable search hits


nrshapiro:
The v4208 client does not yet display full-text with the search hits marked, a core (and key) requirement.  But it occurs to me it may not be obvious to you how to do that.  The server supports this, and does a lot of the work for the client, but being stateless, you need to tell the server the search context again when you request the document by including the previous query parameters in the "search" parameter.  I'll give an example here.

Suppose the user does the search which is submitted to the server as:
    fulltext1="Evenly Suspended Attention"~25

You have the search results, but you cannot get the full-text document from that.  Instead, you need to do a 
   ```/v2/Documents/Document``` request to get the specific document for viewing.

But the stateless server does not remember your query, so when you request any one of the results as full-text, for example:

   ```/v2/Documents/Document/JCP.001.0246A```

you will just get the document, without the hit information.  

Instead, you need to feed the query parameter string from your search URL back to the server via the `search` parameter:

``` /v2/Documents/Document/JCP.001.0246A?search=?fulltext1=%22Evenly%20Suspended%20Attention%22~25&viewperiod=4&formatrequested=HTML&highlightlimit=5&facetmincount=1&facetlimit=15&sort=score%20desc&limit=15```

http://development.org:9100/v2/Documents/Document/JCP.001.0246A/?search=?fulltext1="Evenly%20Suspended%20Attention"~25&viewperiod=4&formatrequested=HTML&highlightlimit=5&facetmincount=1&facetlimit=15&sort=score%20desc&limit=15
e.g.,

```http://development.org:9100/v2/Documents/Document/JCP.001.0246A/?return_format=HTML&search=?fulltext1="Evenly Suspended Attention"~25&viewperiod=4&formatrequested=HTML&highlightlimit=5&facetmincount=1&facetlimit=15&sort=score desc&limit=15```

Note I submitted the entire query parameter list, but the only think I needed here was:

```http://development.org:9100/v2/Documents/Document/JCP.001.0246A/?return_format=HTML&search=?fulltext1="Evenly Suspended Attention"~25&viewperiod=4&formatrequested=HTML&highlightlimit=5&facetmincount=1&facetlimit=15&sort=score desc&limit=15```

and then the document, an excerpt shown here in HTML, will have hits indicated as shown in this excerpt:

[...]

PR #143, opened 2020-12-02
Story/gh72 navigate search hits

### LINKED ISSUE #91: artsub is part of the title
image

image

image

bakerac4:
Fixes #91 and #72

### DIFF
--- a/pep/app/pods/components/document/text/component.ts
+++ b/pep/app/pods/components/document/text/component.ts
@@ -166,6 +166,7 @@ export default class DocumentText extends Component<DocumentTextArgs> {
     replaceHitMarkerText(text: string) {
         let anchorCount = 0;
         let regex = SEARCH_HIT_MARKER_REGEX;
+        let totalAnchorCount = text.match(new RegExp(HIT_MARKER_START, 'g'))?.length ?? 1;
         return text.replace(regex, (match: string) => {
             const { previous, next } = buildJumpToHitsHTML(anchorCount);
             if (match === HIT_MARKER_START) {
@@ -178,7 +179,7 @@ export default class DocumentText extends Component<DocumentTextArgs> {
                     return '';
                 }
             } else if (match === HIT_MARKER_END) {
-                return `${HIT_MARKER_END_OUTPUT_HTML}${next}</span>`;
+                return `${HIT_MARKER_END_OUTPUT_HTML}${anchorCount < totalAnchorCount ? next : ''}</span>`;
             } else {
                 return match;
             }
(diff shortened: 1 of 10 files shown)


pep/app/pods/components/document/text/component.ts
```diff
@@ -242,7 +331,23 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+    }
+
+    /**
+     * Animated the scroll to an element using a specified vertical offset
+     *
+     * @param {(Element | null)} [element]
+     * @memberof DocumentText
+     */
+    animateScrollToElement(element?: Element | null) {
+        const container = this.containerElement?.closest('.page-content-inner');
```
billdami, line 344: i would store this element once on the did-insert of the component instead of finding it each time (probably in the `setupListeners()` action)

pep/app/utils/dom.ts
```diff
@@ -101,3 +102,37 @@ export async function loadXSLT(this: any): Promise<Document | null> {
...
 }
+
+/**
+ * Convert object to browser encoded string
+ *
+ * @export
+ * @param {*} object
+ * @return {*}
+ */
+export function convertToBrowserQueryParamString(object: any) {
```
billdami, line 113: is there a reason we can't use our existing `serializeQueryParams()` util for this?

pep/app/pods/read/document/route.ts
```diff
@@ -41,9 +43,43 @@ export default class ReadDocument extends PageNav(Route) {
      * Fetch the requested document
      * @param {ReadDocumentParams} params
      */
-    model(params: ReadDocumentParams) {
+    model(params: ReadDocumentParams, transition: Transition) {
+        let searchQueryString;
+        if (params.q) {
```
billdami, line 48: why does the search query string only get added if the `q` param (the smart text field) has a value? As i believe its valid for a user to do a search that has that blank, but has specific search term fields (the add-able fields below it in the form) populated.

billdami (changes requested):

pep/app/utils/dom.ts
bakerac4: Yes, they are formatted differently. He basically is expecting us to grab the window/browser query string and use that.

billdami: i guess im just not seeing where the difference in formatting is, as that existing util does browser encoding and stuff as well, is it how it handles arrays or something?

if we need this second util though, it probably makes more sense to live in `utils/url.ts` than `dom.ts`

billdami:
@bakerac4 is this ready for re-review?

bakerac4:
> @bakerac4 is this ready for re-review?

Not yet, Im seeing some issues on his side and he hasn't been able to work on them because pads went down?

PR #148, opened 2020-12-03
Fixes ts errors and removing refine option changing until search hit

### LINKED ISSUE #133: Word Wheel speed impacted by calls to Search with each keypress
Currently, when you type in the fields, it sends both a word wheel search, and a fulltext search to the server.  

Calling the server to get facets with fewer than 4 or 5 characters typed is not really a good use of resources, and it slows down the performance of the word wheel, making it less useful unless you are a hunt and peck typist.

The facets aren't really needed until you either press Search, enter a "space" to start a second word, or pause typing for 1 or 2 seconds.  It's very nice to have the facets once you enter a real word, even if you plan to enter a second or third, but not much use with a partial word.

To experiment with this, I've set the server to return a 422 error if you send a search with less than 5 characters.  I don't think we have **important** words of 4 characters, but really, that's not the right way to implement the restriction anyway.

Thus, I would like you to implement this in the client.  That way, you can send word wheel but not full-text searches for the first 3 or 4 characters, and then if, for whatever reason, the user does enter a three letter word and presses "Search", it will be executed (after I remove the restriction.)  But I wanted to try it to check whether the word wheel works better, and I think it does.

bakerac4:
Fixes #133

### DIFF
--- a/pep/app/pods/most-cited-loading/route.ts
+++ b/pep/app/pods/most-cited-loading/route.ts
@@ -1,7 +1,10 @@
 import Route from '@ember/routing/route';
+
 import { PageNav } from 'pep/mixins/page-layout';
 
 export default class MostCitedLoading extends PageNav(Route) {
     navController = 'most-cited';
-    navTemplate = 'most-cited/nav';
+    get navTemplate() {
+        return 'most-cited/nav';
+    }
 }
--- a/pep/app/pods/application/adapter.ts
+++ b/pep/app/pods/application/adapter.ts
@@ -67,6 +67,7 @@ export default class Application extends DS.RESTAdapter.extend(FastbootAdapter) 
      * When using ember-fetch with ember-simple-auth, authorization headers must be manually set
      * @returns Object
      */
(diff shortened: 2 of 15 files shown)


pep/app/pods/most-cited-loading/route.ts
```diff
@@ -1,7 +1,10 @@
 import Route from '@ember/routing/route';
+
 import { PageNav } from 'pep/mixins/page-layout';
 
 export default class MostCitedLoading extends PageNav(Route) {
     navController = 'most-cited';
-    navTemplate = 'most-cited/nav';
+    get navTemplate() {
+        return 'most-cited/nav';
+    }
```
billdami, line 9: why is this needed for navTemplate but not navController?

bakerac4: It's due to how navTemplate is defined in the function. `navController` is a property and `navTemplate` is a getter

billdami (approved):

PR #163, opened 2020-12-09
disassociated figure

### LINKED ISSUE #121: Expert Pick of the Day - Disassociated Figure
While we originally didn't think it was important that the figure with the picture of the day need be related, and it looks very nice, the problem is that the way it's laid out, they appear to be implied to be related.

We discussed this (DT/NRS/NL) and we want to at least 
1) provide a separating bar between them, and 
2) provide some means to understand this is to a different paper and a means to click the link to the paper.  

Just putting this on the record...we can work it out when we figure out the programming/setup of the Expert Pick of the day, probably from the admin pages.

napafundi:
Fixes #121

### DIFF
--- a/pep/app/pods/components/home/component.ts
+++ b/pep/app/pods/components/home/component.ts
@@ -14,6 +14,7 @@ import ConfigurationService from 'pep/services/configuration';
 import FastbootMediaService from 'pep/services/fastboot-media';
 import PepSessionService from 'pep/services/pep-session';
 import SidebarService from 'pep/services/sidebar';
+import { buildSearchQueryParams } from 'pep/utils/search';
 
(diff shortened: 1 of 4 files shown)


pep/app/pods/components/home/component.ts
```diff
@@ -65,6 +66,44 @@ export default class Home extends Component<HomeArgs> {
...
+    }
+
+    /**
+     * Load the `Document` that contains the expert image
+     * of the day.
+     * Transition to this document's read page.
+     */
+    @action
+    async transitionToImageDocument() {
+        try {
```
bakerac4, line 95: You should use the loading bar here, to provide feedback

PR #165, opened 2020-12-10
Story/gh138 print export in read

### LINKED ISSUE #138: Print / Export Functions in the read pane and the "center toolbar"
In the spec, it was mentioned we should have icons for export and printing in the article pane (Section 2.4 intro) even though they weren't illustrated in the mockups.   In addition, printing articles is mentioned as a core feature in Section 1.2.4, and again mentioned in Section 2.4, but it was not explicitly listed in the subsection requirements (unfortunately).  

Printing/Exporting of the **article** is a core feature and thus we need a print icon (and the underlying functionality) included in the article pane.  Rather than have a toolbar for those at the top, as I was originally, thinking (but as I mentioned in the spec I didn't illustrate that toolbar) you could have the export and print icons integrated into the top of the document as illustrated here.  The export should pull down, to offer the various options for export, which are already available in the staged version, but only when viewing abstracts (remember, some people skip abstracts, per the option).

image

Note that the right click print option seems to work on stage already, but it only prints the first page of the document--I'm not sure why.  The interface to that is very similar to what we already have on PEP-Web, except that works for the entire document, and it prints PEP's the copyright warning message at the bottom, which I'm sure we'll need to do here as well.

One easy option to consider for printing might be when the user clicks the print icon, download the PDF using the server feature, and print that (without making the user save the pdf first, then print).  In a quick test of this though, the formatting of the downloaded PDF is inferior to the displayed version, but it may just be a matter of updating the server's XSLT file with the one you've been updating.

bakerac4:
Fixes #138

### DIFF
--- a/pep/app/pods/read/document/controller.ts
+++ b/pep/app/pods/read/document/controller.ts
@@ -434,7 +434,7 @@ export default class ReadDocument extends Controller {
 
     @action
     async printDocument() {
-        let url = `${ENV.apiBaseUrl}/${ENV.apiNamespace}/Documents/Downloads/PDF/${this.model.id}?client-id=${ENV.clientId}`;
+        let url = `${ENV.apiBaseUrl}/${ENV.apiNamespace}/Documents/Downloads/PDF/${this.model.id}/?client-id=${ENV.clientId}`;
         if (this.session.isAuthenticated && this.session.data.authenticated.SessionId) {
             url += `&client-session=${this.session.data.authenticated.SessionId}`;
         }
--- a/pep/app/pods/read/document/controller.ts
+++ b/pep/app/pods/read/document/controller.ts
@@ -438,19 +438,7 @@ export default class ReadDocument extends Controller {
         if (this.session.isAuthenticated && this.session.data.authenticated.SessionId) {
             url += `&client-session=${this.session.data.authenticated.SessionId}`;
         }
-        fetch(
-            'http://stage-api.pep-web.rocks/v2/Documents/Downloads/PDF/IJP.027.0099A/?client-id=2&client-session=29E7C852-009B-4956-B10E-A019578F074E',
-            {
-                method: 'GET',
-                headers: {
-                    'Content-Type': 'application/pdf'
-                }
-            }
-        ).then((response) => {
-            let localPdf = new window.Blob([response], { type: 'application/pdf' });
-            localPdf = window.URL.createObjectURL(localPdf);
-            printFrame.setAttribute('src', localPdf);
-        });
+        this.printer.printElement(url);
     }
 
     @action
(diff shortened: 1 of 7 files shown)


pep/app/pods/read/document/controller.ts
```diff
@@ -407,6 +431,20 @@ export default class ReadDocument extends Controller {
             }
         });
     }
+
+    @action
+    async printDocument() {
+        let url = `${ENV.apiBaseUrl}/${ENV.apiNamespace}/Documents/Downloads/PDF/${this.model.id}/?client-id=${ENV.clientId}`;
+        if (this.session.isAuthenticated && this.session.data.authenticated.SessionId) {
+            url += `&client-session=${this.session.data.authenticated.SessionId}`;
+        }
```
billdami, line 440: you could make the `downloadAuthParams` getter (which is the cleaner way to serialize url params i think) work for this too by just tweaking the getter to conditionally add the `client-session` param to the object only when they are logged in.

pep/app/services/exports.ts
```diff
@@ -87,17 +88,24 @@ export default class ExportsService extends Service {
             navigator.msSaveBlob(blob, filename);
             return;
         }
+        this._downloadItem(blob, filename, mimetype);
+    }
 
+    _downloadItem(urlOrBlob: string | Blob, filename: string, mimetype?: string) {
```
billdami, line 94: why is this method named with the `_` "im private" convention when its used publicly outside of the class?

also missing docblock comment on this method

pep/app/session-stores/application.js
```diff
@@ -2,7 +2,7 @@ import CookieStore from 'ember-simple-auth/session-stores/cookie';
 import ENV from 'pep/config/environment';
 
 export default CookieStore.extend({
-    cookieName: 'pepweb_session',
+    cookieName: 'pep_session',
```
billdami, line 5: this was changed again? 🤔

pep/app/styles/_pep-documents.scss
```diff
@@ -24,3 +24,17 @@
...
+.pep-abstract-preview {
+    p.title,
+    p.title_author,
+    p.figure,
+    p.pagenumber,
+    p:empty,
+    pb {
+        display: none;
+    }
+}
```
billdami, line 40: you removed this class from the `home` template, but we are adding back the styles? I thought we decided to get rid of this.

billdami (changes requested):

pep/app/styles/_pep-documents.scss
bakerac4: We need it for just the abstract preview widget

pep/app/session-stores/application.js
billdami: ? just  curious as to why this is showing up as a diff again, as i thought this change was already in master from the prior PR.

bakerac4: @billdami Probably because I haven't merged in master. But since there are no conflicts it should just merge together

billdami (approved):

PR #167, opened 2020-12-11
Story/tp65866 color themes

bakerac4:


pep/app/constants/themes.ts
```diff
@@ -57,4 +59,40 @@ export const THEME_TOMATO: Theme = {
...
 };
 
-export default [THEME_DEFAULT, THEME_TOMATO];
+export const THEME_DARK: Theme = {
+    id: ThemeId.DARK,
+    cssPath: '/assets/themes/dark.css',
+    label: 'theme.themes.dark',
+    colors: {
+        links: {
+            glossary: '#a4a29e',
```
billdami, line 68: this is fine for now, if we get the approval to go the SASS-only route for themes, I think it would be good to eventually move these color values into the SASS theme files too as variables. And then create associated classes for them (e.g. `.link-glossary`, etc) which map to those colors in the base SASS files, and finally switch the stuff that uses this to apply those classes instead.

That way all theme related stuff is completely self contained in the SASS, and don't to worry about any less than ideal dynamic style block/inline styles implementations, and any other deps doing that in-JS might incur.

pep/app/templates/head.hbs
```diff
@@ -7,9 +7,7 @@ setting values on the 'head-data' service.
     {{this.model.title}}
 </title>
 <meta name="description" content="A Psychoanalytic Library at your fingertips" />
-{{#if this.model.themePath}}
```
billdami, line 10: i know there was a reason i had this wrapped in an if block when i first did this...but cant remember exactly why now. Maybe there were instances in runloops where the themePath was not defined yet, on boot or something, so the `<link>` would spew 404s?

pep/translations/en-us.json
```diff
@@ -567,7 +567,9 @@
     "theme": {
         "themes": {
             "default": "Navy (default)",
-            "tomato": "Tomato"
+            "tomato": "Tomato",
+            "dark": "Dark",
+            "highContrast": "High Contrast"
         }
```
billdami, line 573: i know one of the reqs was to show an associated description of the theme (maybe in a tooltip or below the preference select input as help text?), did we want to complete that in this PR?

billdami (changes requested):

pep/app/templates/head.hbs
bakerac4: From what I can see, everything works great. Things are a bit different now - as this template only gets run in fastboot

pep/translations/en-us.json
bakerac4: Skipping for now - re @adistasio

pep/app/constants/themes.ts
bakerac4: @billdami agreed

billdami (approved):

PR #174, opened 2020-12-14
feature/tp66767-account-info

### LINKED ISSUE #122: Login/Logout Menu Option disappears
1) I was logged in on an account without full-text access
2) I logged out
3) I tried to log in, but the dialog suddenly disappeared, and then there was no login or logout option anymore.

image

Running in the latest version of Opera.

napafundi:
Fixes #122

### DIFF
--- a/pep/app/pods/components/modal-dialogs/user/info/component.ts
+++ b/pep/app/pods/components/modal-dialogs/user/info/component.ts
@@ -1,6 +1,7 @@
 import CurrentUserService from 'pep/services/current-user';
 
 import { inject as service } from '@ember/service';
+import { action } from '@ember/object';
 import Component from '@glimmer/component';
 
 interface ModalDialogsUserInfoArgs {
@@ -9,4 +10,13 @@ interface ModalDialogsUserInfoArgs {
 
 export default class ModalDialogsUserInfo extends Component<ModalDialogsUserInfoArgs> {
     @service currentUser!: CurrentUserService;
+
+    /**
+     * Open the PaDS home-site using the user's generated
+     * link in a new tab.
+     */
+    @action
+    openPadsTab() {
+        window.open(this.currentUser.user!.paDSHomeURL, '_blank');
+    }
 }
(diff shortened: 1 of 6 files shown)


pep/app/pods/components/modal-dialogs/user/info/component.ts
```diff
@@ -1,12 +1,24 @@
...
     @service currentUser!: CurrentUserService;
+    @service('pep-session') session!: PepSessionService;
+
+    /**
+     * Open the PaDS home-site using the user's generated
+     * link in a new tab.
+     */
+    @action
+    openPadsTab() {
+        window.open(this.currentUser.user!.paDSHomeURL, '_blank');
```
bakerac4, line 22: @napafundi it would probably be safer her to wrap it inside an if block instead of using `!`

pep/app/pods/user/model.ts
```diff
@@ -1,21 +1,46 @@
...
     @attr('boolean') hasCurrentAccess!: boolean;
+    @attr('string') loggedInMethod!: string;
+    @attr('string') paDSHomeURL!: string;
     @attr('date') subscriptionEndDate!: Date;
+    @attr('string') userFullName!: string;
     @attr('string') userName!: string;
     @attr('string') userType!: UserType;
+
+    get activeSubscriptionsJSON(): ActiveSubscription[] {
+        return JSON.parse(this.activeSubscriptions).map((subscription: ActiveSubscription) => {
```
bakerac4, line 36: Is it possible to have no active subscriptions? If so, we should check if an empty array or null gets returned and guard for it

pep/app/pods/components/modal-dialogs/user/info/component.ts
napafundi: I have the button wrapped in an `{{#if this.currentUser.user.paDSHomeURL}}`, that's why I was using the `!` here. Should I still add the if block here too?

pep/app/pods/user/model.ts
napafundi: Yeah this is really weird because it's a string JSON array being returned. I'll add a check to see if it's an empty string/null and default to an empty array either way.

pep/app/pods/components/modal-dialogs/user/info/component.ts
bakerac4: I would, in case you ever change it or try to use this method in another place that doesn't have that.

pep/app/pods/user/model.ts
bakerac4: Sounds good

PR #175, opened 2020-12-14
Story/tp65867 localization

bakerac4:


pep/app/services/theme.ts
```diff
@@ -43,30 +43,59 @@ export default class ThemeService extends Service {
...
+        ];
+    }
+
+    /**
+     * Inject link colors into the linkColors stylesheet. This happens to the administrators can control the colors of the links in the documents
+     *
+     * @memberof ThemeService
+     */
+    @dontRunInFastboot
+    injectLinkColors() {
```
billdami, line 86: @adistasio did we get any feedback/clarification on theme management yet? I'd love if we could get rid of the extra overhead/complexity of this type of stuff, if it all could just live in the SASS.

pep/app/services/theme.ts
```diff
@@ -43,30 +43,59 @@ export default class ThemeService extends Service {
      */
     updateTheme(newThemeId: ThemeId) {
         this.currentUser.updatePrefs({ [PreferenceKey.THEME]: newThemeId });
-        this.setup();
+        if (window.document.querySelector('#theme')) {
```
billdami, line 46: should throw this in a var to eliminate the extra DOM query:
```ts
const themeEl = window.document.querySelector('#theme');
if (themeEl) {
    themeEl.setAttribute(....)
}
```

pep/app/services/lang.ts
```diff
@@ -14,21 +16,36 @@ export default class LangService extends Service {
...
-    setup() {
+    async setup() {
         // TODO eventually possibly allow setting via ?lang={code} app-level query param
         this.currentLanguage = this.currentUser.preferences?.lang ?? LANG_EN_US.code;
+        await this.loadLanguage(this.currentLanguage);
         this.intl.setLocale(this.currentLanguage);
     }
 
+    async loadLanguage(lang: LanguageCode) {
+        const translations = await fetch(`/translations/${lang}.json`);
```
billdami, line 37: we should do this, to allow for optimal browser caching/cache-busting of the translation files:

https://ember-intl.github.io/ember-intl/versions/master/docs/guide/asynchronously-loading-translations#fingerprinting

should be pretty straightforward, I'd probably just store the full uninterpolated paths in the lang object constants themselves, so that the fingerprinting replacement happens automatically e.g. 

```ts
export const LANG_FR_FR: Language = {
    code: LanguageCode.frFR,
    label: 'language.codes.frFR',
   path: 'assets/translations/fr-fr.json'
};
```
I think we need the fingerprinting rewriting of the url paths to happen in any case, because we'll want the translations files served from the S3 CDN, not the local fastboot server. The fingerprinting process prepends that assets base url for you as well.

billdami (changes requested):

billdami (approved):

PR #182, opened 2020-12-16
fix login button appearing when IP logged in

napafundi:


pep/app/pods/components/page/drawer/template.hbs
```diff
@@ -112,30 +112,30 @@
...
-                                    type="button"
-                                    class="list-group-item list-group-item-action"
-                                    {{on "click" this.openLoginModal}}
-                                >
-                                    <FaIcon @icon="sign-in" @fixedWidth={{true}} class="mr-2" />
-                                    {{t "nav.user.login"}}
-                                </button>
-                            </Page::Drawer::NavItem>
+                        {{#if this.session.canLogOut}}
+                            {{#if this.session.isAuthenticated}}
```
bakerac4, line 116: Im not sure the `else` here is reachable due to this all being inside of an `if` already that checks if your authenticated

napafundi:
@bakerac4 updates in.

PR #187, opened 2020-12-17
Story/tp65869 intro mode

bakerac4:


pep/app/styles/themes/high-contrast.scss
```diff
@@ -20,6 +20,16 @@ $color-backdrop: #fff;
...
 $color-focus-shadow: rgba($color-brand-primary, 0.25);
+
+$document-link-glosstip-hover: #03a678;
+$document-link-glosstip-bg: rgba($document-link-glosstip-hover, 0.1);
+$document-link-bibliography: #12130f;
+$document-link-page: #1d84b5;
+$document-link-figure: #71697a;
+$document-link-external: #a98743;
+$document-link-webExternal: #f7c548;
+$document-link-webInternal: #305252;
```
billdami, line 31: why the mix of kebab and camel casing on these vars?

pep/app/services/theme.ts
```diff
@@ -3,10 +3,8 @@ import Service, { inject as service } from '@ember/service';
 import HeadDataService from 'ember-cli-head/services/head-data';
 import IntlService from 'ember-intl/services/intl';
 
-import Color from 'color';
```
billdami, line 6: can we remove this `color` package dep now?

pep/app/pods/components/document/text/component.ts
```diff
@@ -261,6 +257,25 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+            const parent = target.parentElement;
+            if (parent) {
+                const sourceCode = parent.getAttribute('data-journal-code');
+                const volume = parent.getAttribute('data-volume');
+                const page = parent.getAttribute('data-page');
+
+                if (sourceCode && volume) {
+                    this.router.transitionTo('browse.journal.volume', sourceCode, volume);
+                }
+            }
```
billdami, line 278: I think i mentioned this in an older PR but this onDocumentClick action and if block is just getting ginormous. Would be nice if we could chunk this out into like a hash that maps to a bunch of separate helper methods, e.g. something like

```ts
const documentClickActions = {
   [DocumentLinkTypes.GLOSSARY_TERM]: 'viewGlossaryTerm',
   [DocumentLinkTypes.AUTHOR_SEARCH]: 'viewAuthor',
   ...
}
```
and then this method just gets reduced to mainly just calling the correct method from that hash essentially:

```ts
onDocumentClick(event: Event) {
    //common stuff here...
   this[documentClickActions[type]](target);
}
```

pep/app/pods/application/route.ts
```diff
@@ -132,6 +136,119 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
...
+        const steps: Step[] = [];
+        if (this.media.isMobile || this.media.isTablet) {
+            steps.push({
+                id: 'home',
+                attachTo: {
+                    element: '.navbar .navbar-toggler',
+                    on: 'bottom'
+                },
+
+                text: 'This button takes you home',
```
billdami, line 167: untranslated text

pep/app/pods/application/route.ts
```diff
@@ -132,6 +136,119 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
...
+                ]
+            });
+        } else {
+            steps.push({
+                id: 'home',
+                attachTo: {
+                    element: '.nav-item.home',
+                    on: 'auto'
+                },
+                text: 'This button takes you home',
```
billdami, line 183: same as above ^

pep/app/pods/application/route.ts
```diff
@@ -132,6 +136,119 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
...
+        }
+        steps.push(
+            {
+                id: 'toggle-left',
+                attachTo: {
+                    element: '.sidebar-left .sidebar-toggle-handle',
+                    on: 'auto'
+                },
+                classes: 'tour-left-sidebar-spacing',
+                text: 'This button shows/hides the search',
```
billdami, line 201: untranslated text here too (and below)

pep/app/pods/application/route.ts
```diff
@@ -132,6 +136,119 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
...
+    /**
+     * If the tour is enabled, show it
+     *
+     * @param {ApplicationController} controller
+     * @param {*} model
+     * @memberof Application
+     */
+    async setupController(controller: ApplicationController, model: any) {
+        super.setupController(controller, model);
+        if (this.currentUser.preferences?.tourEnabled) {
```
billdami, line 247: should we skip executing this in fastboot? (even if the tour itself will stop itself from executing in fastboot, might be good to just short-circuit this here)

billdami (changes requested):

pep/package.json
```diff
@@ -65,7 +64,6 @@
         "babel-eslint": "^10.1.0",
         "bootstrap": "4.5.1",
         "broccoli-asset-rev": "^3.0.0",
-        "color": "^3.1.3",
```
billdami, line 67: @bakerac4 did you use `yarn remove` to remove this dep? doesnt look like there were any changes to the yarn.lock in this push.

PR #195, opened 2020-12-21
Fixing issue with search terms not saving

### LINKED ISSUE #189: Search fields are not restored when using "find terms or phrases" instead of smarttext...
**Case 0: (working)**
1) Do a search with smarttext, e.g. cardiac arrest
2) Go to results
3) Click on Bergstein (2015).  Go to read (if you are not already there via skip abstracts)
4) Use Back to search link.  Search args are restored

**Case 1:(not working)**
1) Do the same search with field "articles" (not in smarttext)
2) Go to results
3) Click on Bergstein (2015).  Go to read (if you are not already there via skip abstracts)
4) Use Back to search link.  Search is not restored.

**Note for Case 1, the search parameter in the URL in documents/document is:**

'facetfields=art_year_int,art_views_last12mos,art_cited_5,art_authors,art_lang,art_type,art_sourcetype,art_sourcetitleabbr,glossary_group_terms,art_kwds_str&facetlimit=15&facetmincount=1&highlightlimit=4&synonyms=false&fulltext1=¶scope=doc'

**In Case 0, the search parameter is** 

'facetfields=art_year_int%2Cart_views_last12mos%2Cart_cited_5%2Cart_authors%2Cart_lang%2Cart_type%2Cart_sourcetype%2Cart_sourcetitleabbr%2Cglossary_group_terms%2Cart_kwds_str&facetlimit=15&facetmincount=1&highlightlimit=4&synonyms=false&fulltext1=cardiac+arrest'

Note in Case 1, fulltext1 param empty
   And I'm not sure what  the arg scope is?  I note in diagnosing this that the client is at least sometimes sending parascope to search (e.g., set field to biblio).  But parascope only applies to parameter paratext, which you are not using.  

Note in Case 0, search words are restored correctly into smarttext, even though the search= shows them in fulltext1.  There may be other undetected incorrect consequences of it being passed in search as fulltext1 instead of smarttext.  I didn't investigate that.

bakerac4:
Fixes #189

### DIFF
--- a/pep/app/pods/read/document/route.ts
+++ b/pep/app/pods/read/document/route.ts
@@ -18,8 +18,6 @@ import SidebarService from 'pep/services/sidebar';
 import { buildSearchQueryParams, copyToController, hasSearchQuery } from 'pep/utils/search';
 import { serializeQueryParams } from 'pep/utils/url';
 
-import { copySearchToController } from '../../../utils/search';
-
 export interface ReadDocumentParams {
     document_id: string;
     q: string;
--- a/pep/app/pods/read/document/route.ts
+++ b/pep/app/pods/read/document/route.ts
@@ -18,6 +18,8 @@ import SidebarService from 'pep/services/sidebar';
 import { buildSearchQueryParams, copyToController, hasSearchQuery } from 'pep/utils/search';
 import { serializeQueryParams } from 'pep/utils/url';
 
+import { copySearchToController } from '../../../utils/search';
+
 export interface ReadDocumentParams {
     document_id: string;
     q: string;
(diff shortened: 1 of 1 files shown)


pep/app/pods/read/document/route.ts
```diff
@@ -18,6 +18,8 @@ import SidebarService from 'pep/services/sidebar';
 import { buildSearchQueryParams, copyToController, hasSearchQuery } from 'pep/utils/search';
 import { serializeQueryParams } from 'pep/utils/url';
 
+import { copySearchToController } from '../../../utils/search';
```
billdami, line 21: looks like an unused import

billdami (changes requested):

billdami (approved):

PR #202, opened 2020-12-22
Theme fixes

bakerac4:


pep/app/styles/themes/dark.scss
```diff
@@ -16,9 +16,8 @@
 $logo-path: '/assets/images/logo-dark.jpg';
 $color-brand-primary: #2f3136;
 $color-backdrop: #36393f;
-// $color-backdrop: #474954;
-// $color-primary-background: #545763;
 $color-primary-background: #393c43;
+$color-tooltip-background: darken($color-primary-background, 5%);
```
billdami, line 20: you might've already caught this but i also noticed on the last stage build that the dropdown menu divider (like between "About" and "Login" in the user menu) bg color was incorrect in the dark mode theme. I think theres a bootstrap var you can override for that: https://github.com/twbs/bootstrap/blob/v5.0.0-alpha2/scss/_variables.scss#L934

billdami (changes requested):

billdami (approved):

PR #203, opened 2020-12-22
Green builds ♻️

billdami:
@bakerac4 @napafundi One of my goals for the new year is to try to make a concerted push towards making sure our builds are always error/warning free and all tests passing before we merge into master/releases. As we've seen in the past, ignoring build errors/test failures can lead to some very hard to diagnose bugs. 

This is a first step towards that, I got it _very_ close to a perfect build log, sans the one collision warning that we know we don't care about (but still prob should eventually address):
Screen Shot 2020-12-22 at 1 39 01 PM

If we could working that into our local dev flow of checking those right before we open a PR for review, I think that would help alot. The next step would naturally be looking at how to re-enable tests in CI (and hopefully blocking PRs until everything passes). Lastly, I think it'd be great if we could eventually have meaningful unit/integration tests 😄 -- but I know thats more of a stretch goal that will require some broader internal discussion as to what that means/to what extent it would go.


billdami:
@napafundi @bakerac4 can one of you guys review this today? I'd like to get this in sooner than later to establish a clean baseline for the release. To adam's point about not not remembering to check builds/tests for every branch, I think for now maybe we can just do this manual check before each _release_ branch is merged, I don't think that's too big of an ask. Thoughts?

@bakerac4 also I know you don't love that many of the tests are reduced down to simple/non-meaningful `assert.ok()` checks, but I do think this is an improvement over having already non-meaningful failing tests. Mainly because the deluge of failing tests can obscure actual issues (and I did fix a few of the tests for real). Also, while their testing coverage is garbage, these simple `assert.ok()` tests do actually have _some_ value, they are validating that components run and render successfully w/o fatal errors, so having them is still better than nothing at all (even though its not _that_ much better).

pep/app/services/head-data.js
```diff
@@ -2,5 +2,5 @@ import Service from '@ember/service';
 import { tracked } from '@glimmer/tracking';
 
 export default class HeadDataService extends Service {
-    @tracked themePath?: string;
+    @tracked themePath;
```
napafundi, line 5: Is there a reason you're removing the type here?

billdami: its a .js file now 🙈 i converted this service from JS to TS because its overriding a js service from an addon and was giving one of those collision warnings we get. And in this case it was warranted to fix that, because its possible for a build to randomly choose the addon's version over ours left as-is.

napafundi: Oh whoops didn't notice that 😄 yeah the collisions are scary business. Good to go 👍

billdami:
@napafundi could you also just pull latest 006 and re-yarn/run it locally just to sanity check the build for me when you get a sec?

napafundi:
@billdami looks good, just the html-safe collision warning

billdami:
@napafundi thanks! yeah, i mean i'd probably be fine with just converting that html-safe helper to js for now if you want to in your current/next branch. Its not imported/used directly in any other TS files so its not really adding much benefit to warrant ignoring that collision warning. 

_Hopefully_ ember-cli-typescript will provide a better solution to this eventually so we dont need any ugly `.js` files at all tarnishing our codebase 🤞

PR #209, opened 2021-01-04
Adding check for url query param session id

bakerac4:


pep/app/authenticators/ip.ts
```diff
@@ -19,7 +19,11 @@ export default class IpAuthenticator extends CredentialsAuthenticator {
             const sessionData = this.session.getUnauthenticatedSession();
             let url = `${ENV.authBaseUrl}/Authenticate/ip`;
 
-            if (sessionData?.SessionId) {
+            // check for session id in query params
+            if (this.fastboot.request.queryParams.sessionId) {
```
billdami, line 23: is `queryParams` always a defined object, even if the request contained no query params? just making sure we don't need a `?` optional chain here.

bakerac4: @billdami it sure is! Its just an empty object int he request object

billdami (approved):

PR #211, opened 2021-01-04
Translation from api

bakerac4:


pep/app/pods/components/document/text/component.ts
```diff
@@ -467,5 +489,52 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+                            instance._loaded = false;
+                        },
+                        onShow: (instance: DocumentTippyInstance) => {
+                            if (instance._isFetching || instance._loaded) {
+                                return;
+                            }
+
+                            this.args
+                                .loadTranslation?.(paraLangId, paraLangRx)
+                                .then(async (text: string) => {
```
billdami, line 518: could you make the `onShow` function `async` as well, and flatten out these promise calls completely?

pep/app/pods/read/document/controller.ts
```diff
@@ -434,6 +434,23 @@ export default class ReadDocument extends Controller {
         });
     }
 
+    @action
+    async loadTranslation(paraLangId: string, paraLangRx: string) {
```
billdami, line 438: can this just live directly in the `<Document::Text>` component?

billdami (changes requested):

pep/app/pods/components/document/text/component.ts
bakerac4: No, this is typed from the package and it's not a promise. So to keep your GreenBuilds 🟢 we need to keep as is

billdami (approved):

PR #213, opened 2021-01-06
Story/tp65882 preferences

bakerac4:


pep/app/services/current-user.ts
```diff
@@ -255,4 +268,87 @@ export default class CurrentUserService extends Service {
...
+     */
+    setFontSize(newSize: FontSizes) {
+        const document = this.document;
+        let target = document.documentElement;
+        const size = AvailableFontSizes.find((item) => item.id === newSize) ?? FONT_SIZE_DEFAULT;
+        AvailableFontSizes.map((item) => item.class).forEach((item) => {
+            removeClass(target, item);
+        });
+
+        addClass(target, size.class);
```
billdami, line 286: we should be able to remove the need for this imperative manipulation of the document classes by utilizing ember-body-class2, which is already a dep of the project and is fastboot-safe as well.

I believe it allows dynamic updating of the class via getters, if so this could be reduced to just adding the following on the `application` route:
```ts
get classNames() {
    //get the current fontsize from the currentUser service here...
    return currentFontSize.class
}
```

Note though, that this approach may or may not be viable if we need to set the styles on the `<html>` element instead like i mention in my other comment regarding the font-size styles. But this declarative approach _would_ be much nicer.

pep/app/constants/text.ts
```diff
@@ -0,0 +1,76 @@
+export enum FontSizes {
+    XS = 'xs',
+    SMALL = 'small',
+    DEFAULT = 'default',
+    LARGE = 'large',
+    XLARGE = 'xlarge'
+}
```
billdami, line 7: i think we may have strayed from this, but the project was originally following the convention of naming enums in the singular form, e.g. `FontSize`, looks like most of the enums follow that. Imo it just reads a bit better when using them in typing args, etc, as like `(newSize: FontSizes)` reads as if you are passing it multiple font sizes.

pep/app/pods/components/document/text/template.hbs
```diff
@@ -1,5 +1,11 @@
 <div
-    class="document-text"
+    class="document-text
+
+
+        {{if this.currentUser.preferences.glossaryFormattingEnabled "glossary-highlight"}}
+
+
+        {{this.currentUser.textJustification.class}}"
```
billdami, line 8: i'd say its 95% likely they'll want the text justification setting to apply to anywhere we display abstracts too (document preview pane, expert pick of the day, etc), does this apply to those instances?

also why all the extra whitespace here?

pep/app/services/current-user.ts
```diff
@@ -255,4 +268,87 @@ export default class CurrentUserService extends Service {
...
+        this.updatePrefs({ [PreferenceKey.TEXT_JUSTIFICATION]: textJustification });
+    }
+
+    /**
+     * Available text justification options
+     *
+     * @readonly
+     * @memberof CurrentUserService
+     */
+    get availableTextJustifications() {
```
billdami, line 315: can we move the getter declarations above the class methods, to keep in line with our formatting conventions

billdami (changes requested):

pep/app/styles/_typography.scss
```diff
@@ -26,3 +26,19 @@
...
+    font-size: small;
+}
+
+.font-lg {
+    font-size: large;
+}
+
+.font-xlarge {
+    font-size: x-large;
+}
```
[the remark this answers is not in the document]
bakerac4: You are correct, however that's what is currently happening already.

pep/app/services/current-user.ts
bakerac4: Correct, I originally tried that route but it needed to be on the HTML element, not the body.

pep/app/pods/components/document/text/template.hbs
bakerac4: We don't use the document text component in the document preview pane (I think we probably should, but haven't gotten to looking at it yet). It applies everywhere else though.

The whitespace is from my auto formatter. Every time I fix it, it gets added back in

pep/app/services/current-user.ts
billdami: 👍  gotcha yeah i missed the fact that the document element referred to the HTML element, not the body.

ideally, it would be nice if we could override the ember-body-class2 code to allow the html element to be the target instead of `<body>` here: https://github.com/nathanhammond/ember-body-class/blob/master/addon/utils/body-class.js#L68  so that we can keep our app code cleaner and declarative, but i guess we can punt on that for now.

bakerac4: Yes I originally tried to use that, but when I did I was getting build errors that the package didn't exist. I played with it for a bit before giving up. Figured we can iterate later if need bo

pep/app/pods/components/document/text/template.hbs
billdami: thats weird, i can see adding a single linebreak/space between them, by why two. 🤔

and ok, in that case maybe we can just apply the `{{{{this.currentUser.textJustification.class}}` class directly to the preview pane component's abstract content container element?

pep/app/constants/text.ts
```diff
@@ -0,0 +1,76 @@
...
+    class: 'text-center'
+};
+
+export const TEXT_RIGHT: TextJustification = {
+    id: TextJustificationId.RIGHT,
+    label: 'text.justification.right',
+    class: 'text-right'
+};
+
+export const TextJustifications = [TEXT_LEFT, TEXT_CENTER, TEXT_RIGHT];
```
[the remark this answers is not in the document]
billdami: @nrshapiro can you provide clarity here, and confirm whether my above comment is correct, regarding this requirement?

```
[Core] ​The user shall be able to select whether the text is left-right justified or ragged.
```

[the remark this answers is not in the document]
nrshapiro: @billdami 

If it cannot be added easily, please defer this requirement until a later release.  It's not essential from the start.  There are more important issues.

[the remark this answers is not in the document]
billdami: @nrshapiro the feature is already implemented, we just are looking for a clarification on which justification style options you wanted, which would take virtually no time to adjust. Are you looking for: left-aligned (right side ragged), right-aligned (left side ragged), and center-aligned (both sides ragged), or  left-aligned (right side ragged) and justified (no ragged edges), or a combination of any of those?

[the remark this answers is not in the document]
billdami: @nrshapiro merging what we have for now, but just response this when you can, so we can quickly adjust with the correct options, if needed.

Issue #50, opened 2020-09-30
Highlighting Search Matches (was 'Missing Requirements')


nrshapiro:
Sorry...I think I left out a few important (related) requirements for the read pane, at least as explicit numbered items.

As it says in the core features on the bottom of page 6: 

- Highlight search matches, and provide navigation among them, e.g., Prev/Next intra-document match, Prev/Next interdocument match

I don't think these are in the list of Reading Pane requirements (or at least I'm not seeing them now.)  But this is an import CORE feature.

Here it is in PEP-Web: (arrows around hits, arrows in toolbar).

![image](https://user-images.githubusercontent.com/4837554/94645262-657f9f00-02b9-11eb-81a6-cdb83bcaf9ed.png)

bakerac4:
@nrshapiro can you explain this in more detail. 

For example, do the arrows on top do both intra-document and inter-document matching? Or if you click on the right arrow up top do you just go to the next document in the list.

nrshapiro:
@bakerac4 

The single arrows on top go to next and previous hit (Prev/Next intra-document match)

The double arrows go to the previous and next document and first hit (Prev/Next interdocument match)

The illustration only shows the "next" interdocument arrow because this is the first document in the match set, and they show the arrows dynamically.

nrshapiro:
@bakerac4   I see this is working, but I have some comments about that.

1) Please use graphical arrows.  The < > are hard to see/find when scanning the text.  The ones used by the server when it generates HTML look much better.  E.g., 🡄 and 🡆  These are just unicode characters.

2) Would it be easy enough to also link the "hits" listed in the read pane to the hits in the article?  That would be a nice way to see those listed in the left pane directly in context.

nrshapiro:
This issue needs to be elevated (it was already marked high priority, but I will lock it at the top of the issues since it's fallen on the second page now): it was in fact raised again during prod site preview testing by a board member.  I will bring it up in the tech meeting but here's what is needed, and a clarification of note 2 in my message from 24 days ago above.

1) There needs to be a way to easily navigate hits without actually scrolling to find them.  Per the spec, this is specified (though not illustrated) to be on the toolbar, per our existing site.  Picture from existing site:

![image](https://user-images.githubusercontent.com/4837554/103549250-0f506800-4e75-11eb-8c1d-eee738743f68.png)

[...]

PR #216, opened 2021-01-07
Story/gh50 search matching

bakerac4:


bakerac4:
Fixes #50

pep/app/pods/read/document/template.hbs
```diff
@@ -33,36 +33,79 @@
...
+                                @action={{this.viewPreviousSearchHit}}
+                                @type="link"
+                                @icon="chevron-left"
+                                @iconPrefix="fas"
+                                @iconSize="lg"
+                                @tooltip={{t "common.previous"}}
+                                class="py-0 px-1 mx-1"
+                            />
+                        {{/if}}
+                        {{this.q}}
```
billdami, line 49: do search hits only come from "smart search"-based queries (which is what `q` holds), or can hits also occur from fulltext search term-based hits? If its the latter, i think you'd need a bit more logic here to build this displayed text value

billdami (changes requested):

billdami (approved):

PR #230, opened 2021-01-12
Feature/tp65871-report-data-error

napafundi:
https://gavant.tpondemand.com/entity/65871-report-a-data-error-dialog


pep/app/pods/components/modal-dialogs/help/report-data-error/component.ts
```diff
@@ -0,0 +1,126 @@
...
+export default class ModalDialogsHelpReportDataError extends Component<ModalDialogsHelpReportDataErrorArgs> {
+    @service ajax!: AjaxService;
+    @service currentUser!: CurrentUserService;
+    @service loadingBar!: LoadingBarService;
+    @service notifications!: NotificationService;
+    @service intl!: IntlService;
+    validations = REPORT_DATA_ERROR_VALIDATIONS;
+    dataErrorUrl = `${ENV.reportsBaseUrl}/data-errors`;
+
+    supportResources: SupportResource[] = [
```
bakerac4, line 48: We usually separate stuff like this out into a constants file

.env-staging
```diff
@@ -15,4 +15,5 @@ COOKIE_DOMAIN=
 COOKIE_SECURE=
 COOKIE_SAME_SITE=
 GOOGLE_ANALYTICS_ID=
-GOOGLE_ANALYTICS_ENVIRONMENT=
\ No newline at end of file
+GOOGLE_ANALYTICS_ENVIRONMENT=
+REPORTS_API=
```
bakerac4, line 19: @napafundi I believe you need to add this to the `dotenv.js` file as well

pep/app/pods/components/modal-dialogs/help/report-data-error/component.ts
napafundi: @bakerac4 yeah the reason I didn't put these entire objects in a constants file is because we already have a `url` constants file that these objects use and I didn't want to necessarily tie a specific "label" to the url (so you could more readily re-use the urls w/ different link labels, etc)

bakerac4: @napafundi I think that you can do both as long as you just name things appropriately. So if you were to create an export called `SUPPORT_RESOURCES` in the `urls.ts` file and export what you have currently you could still not tie the labels in.

Also I just noticed this now, but these need to be translated. Check out `themes.ts` for how we normally do that.

napafundi: @bakerac4 ahh yep, good catch. Yeah I see, I was thinking of making a whole new file but this makes more sense.

PR #242, opened 2021-01-18
Changing logo loading back to API

### LINKED ISSUE #238: Banner Graphic Refresh Needed
I noticed in your repository you have an assets folder with a copy of the banner and gui graphical elements. That's fine (in fact I had suggested it.) However, I've adjusted some of the names in the main storage location (S3) to handle where there were inconsistencies and your code was not referencing them as they were defined.  For example LU-AM:

image

Some of the original names had initial caps (B rather than b) or other case inconsistencies, and I've gone through and fixed these in the S3 repository.  But they have not been refreshed in yours, so they don't show up.

LU-AM
OPUS
REVAPA
GW
SE

Perhaps as part of the data build, the S3 pattern banner.*.gif should be copied to your assets folder?  That way, in the future, as we add journals, or on the rare occaision we need to adjust a banner, we get these copied over to the assets folder.  I am still in favor of keeping a copy that doesn't have to be loaded from S3 with each article as you've done, we just need a way to do an update without rebuilding the app.  Or please tell me the best way for me to do this, as we move forward in the future.

bakerac4:
Fixes #238


pep/app/pods/browse/journal/volume/route.ts
```diff
@@ -21,12 +24,24 @@ export default class BrowseJournalVolume extends Route {
...
+    /**
+     * Load the first journal for a specific source for information
+     *
+     * @memberof BrowseJournalIndex
+     */
+    async afterModel() {
+        const journalParams = this.paramsFor('browse.journal') as BrowseJournalParams;
+        const journal = await this.store.query('journal', { sourcecode: journalParams.pep_code });
+        this.journalInformation = journal.firstObject;
+    }
```
billdami, line 36: both `browse.journal.index` and `browse.journal.volume` fetch the journal model, so we should probably instead fetch it as the parent `browse.journal` route's main `model()` and then in both sub routes just use `this.modelFor('browse.journal')` (which you could do directly in setupController since its synchronous)

billdami (changes requested):

billdami (approved):

Issue #255, opened 2021-01-27
Freud SE/GW concordance (translation) popups partly hidden at top of Window


nrshapiro:
These should be all visible.  Consider the following...same paragraph, clipped quite a bit to only a little depending on where it is in the window. 

<img width="1206" alt="2021-01-27_08h47_07" src="https://user-images.githubusercontent.com/4837554/105999992-5567a880-607c-11eb-81ec-ada4706cb526.png">
<img width="1204" alt="2021-01-27_08h46_50" src="https://user-images.githubusercontent.com/4837554/106000003-58629900-607c-11eb-86f9-5dddea44bbc6.png">
<img width="1187" alt="2021-01-27_08h46_36" src="https://user-images.githubusercontent.com/4837554/106000012-5ac4f300-607c-11eb-8d4d-af960fa55f9b.png">

Issue #254, opened 2021-01-26
Glossary Displays 2 or more Duplicate Sections


nrshapiro:
Clicking on glossary entries is showing a significant problem in display.  It's showing multiple sections of the same definition blocks.

Examples:
<img width="608" alt="2021-01-26_12h55_28" src="https://user-images.githubusercontent.com/4837554/105904071-af219180-5fee-11eb-9312-ca5096a96380.png">

<img width="619" alt="2021-01-26_12h55_12" src="https://user-images.githubusercontent.com/4837554/105904317-fdcf2b80-5fee-11eb-8d1b-5baffebc2a94.png">

<img width="589" alt="2021-01-26_12h56_38" src="https://user-images.githubusercontent.com/4837554/105904208-da0be580-5fee-11eb-93f1-e6394da5016c.png">

In addition, some lines extend beyond the bounds of the glossary popup.

<img width="840" alt="2021-01-26_12h56_05" src="https://user-images.githubusercontent.com/4837554/105904264-eb54f200-5fee-11eb-8f9d-2e8e47347d99.png">

nrshapiro:
@bakerac4 
Note: This seems to only happen when it's called from the Glossary Terms Infocard, rather than the link in documents.

bakerac4:
@nrshapiro it seems like the it works that way due to not having ids. In the document text we have an id to pass usually.

In the glossary terms infocard we call 

https://stage-api.pep-web.rocks/v2/Documents/Glossary/siblingrivalry/?return_format=XML&termidtype=Group

What is the correct call to make to not get dupes?

nrshapiro:
@bakerac4 

I've changed the way the endpoint works which should fix the problem.  The old way it worked was bringing back each term in the group as a new record, and the terms fields were supposed to be populated with the individual term data...but not all term fields were being populated.  Now, I've added a new parameter to the endpoint, to preserve that way, but the default is now to only return one record per matching group.  In my local tests, with that as the default, the client works correctly for groups that previously returned multiple entries.  The reason for the prior duplication is that despite the term fields changing (the ones that were populated anyway), the xml of the group is still the whole group.

[...]

Issue #239, opened 2021-01-15
PEP Current Misleading Message


nrshapiro:
I am logged in, and looking at a PEPCurrent document (embargoed because it's too new).  I see the following:

![image](https://user-images.githubusercontent.com/4837554/104742801-e6539100-5718-11eb-9c78-fbe5f1d872ad.png)

1) I am already logged in, so logging in won't do me any good.
2) In these cases, the second line about logging in should not be shown.

bakerac4:
@nrshapiro are you sure you were logged in? This is what I see when Im logged in

![Screen Shot 2021-01-27 at 9 31 44 AM](https://user-images.githubusercontent.com/9952936/106005787-8ea31700-6082-11eb-9cb5-0bf83dbaa2a5.png)

nrshapiro:
@bakerac4 

Are you logged in as test4?  That has access to current.  The ID I was logged in as does not.  Try one without PEPCurrent access.  2018 and above for that journal.

Issue #220, opened 2021-01-08
Glossary display outside of the popup box


nrshapiro:
Client: 2021.01.06-5647
Server: 8.6.2/2021.0104.2.Alpha
DataSource: OPAS.AWSProd.2021-01-05
Browser: Chrome

Replicate
1) Article (doesn't matter) but: https://stage.pep-web.rocks/read/AOP.038.0052A
2) Click on a highlighted Unconconcious to popup the glossary

I was testing/browsing and I ran across this problem with a glossary entry exceeding the bounds of the popup (by a lot).  Check towards the bottom of the display and note the line that goes all the way to the right, outside of the popup.

![image](https://user-images.githubusercontent.com/4837554/104040976-522c7b80-51a6-11eb-80f4-1d6824a93446.png)

Might be easier to see here:

![image](https://user-images.githubusercontent.com/4837554/104041261-c36c2e80-51a6-11eb-89ae-0e7830c8bb23.png)

PR #258, opened 2021-01-28
Multiple fixes

### LINKED ISSUE #256: Hit Summary - Link at top
1) The hit summary/link at top always has the words munged together.

image

2) If there's only one hit, and you go to it, then scroll around the document, the link is no longer available at the top:

image

bakerac4:
Fixes #256, #255, Part of #254, #239, #220

@nrshapiro in order to fully fix 254, I need more information. See https://github.com/Psychoanalytic-Electronic-Publishing/PEP-Web-User-Interface/issues/254#issuecomment-768350239

### DIFF
--- a/pep/app/pods/search/read/controller.ts
+++ b/pep/app/pods/search/read/controller.ts
@@ -218,6 +218,18 @@ export default class SearchRead extends Controller {
             this.searchHitNumber = this.searchHitNumber - 1;
         }
     }
+
+    /**
+     * Set the search hit number to the next number. Most likely its the same number they were on, so make it undefined first
+     *
+     * @param {number} number
+     * @memberof SearchRead
+     */
+    @action
+    viewSearchHitNumber(number: number) {
+        this.searchHitNumber = undefined;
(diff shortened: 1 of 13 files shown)


pep/app/pods/search/read/controller.ts
```diff
@@ -218,6 +218,18 @@ export default class SearchRead extends Controller {
...
+    /**
+     * Set the search hit number to the next number. Most likely its the same number they were on, so make it undefined first
+     *
+     * @param {number} number
+     * @memberof SearchRead
+     */
+    @action
+    viewSearchHitNumber(number: number) {
+        this.searchHitNumber = undefined;
+        this.searchHitNumber = number;
```
billdami, line 231: I dont _think_ this will have any actual effect on ember/glimmer, as it waits until the end of the runloop to coalesce all value changes before a new render. its more likely in these cases you want to do something like:
```ts
this.searchHitNumber = undefined;
next(this, () => this.searchHitNumber = number);
```

billdami (approved):

PR #260, opened 2021-01-28
login for IP when settings change

### LINKED ISSUE #257: Incorrectly Labeled as Logged In Via IP
@billdami @adistasio 
For Info: @nrshapiro 
Sometimes the client is marking the session as logged in via IP when it isn't. It is hard to pin down what is causing this as it seems to happen in different circumstances. One instance I have just experienced is is opening with a logged on session for user:Test7:
https://stage.pep-web.rocks/?sessionId=43cfd03f-5dde-4428-9d99-42d15e6b6a52

The client opens with the correct user but it is saying logged in via IP which means you can't logoff, without clearing cookies.  The info returned by the PaDS endpoints for user and IP is correct and does not mention Logged on via IP.
Screenshot 2021-01-27 at 16 50 00
Screenshot 2021-01-27 at 16 50 28

bakerac4:
Fixes #257

### DIFF
--- a/pep/app/pods/application/template.hbs
+++ b/pep/app/pods/application/template.hbs
@@ -15,3 +15,4 @@
 <Page::Drawer @openAboutModal={{this.openAboutModal}} />
 <NotificationContainer @position="top" />
 <ModalOutlet />
+<InformationBar class="app-information-bar" @light={{true}} />
--- a/pep/app/pods/application/template.hbs
+++ b/pep/app/pods/application/template.hbs
@@ -15,4 +15,4 @@
 <Page::Drawer @openAboutModal={{this.openAboutModal}} />
 <NotificationContainer @position="top" />
 <ModalOutlet />
-<InformationBar class="app-information-bar" @light={{true}} />
+<InformationBar class="app-information-bar" />
--- a/pep/app/pods/components/information-bar/component.ts
+++ b/pep/app/pods/components/information-bar/component.ts
@@ -1,0 +1,40 @@
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+import Component from '@glimmer/component';
+
+import InformationBarService from 'pep/services/information-bar';
+import { slideDown } from 'pep/utils/animation';
+
+interface InformationBarArgs {
+    light: boolean;
+}
+
+export default class InformationBar extends Component<InformationBarArgs> {
+    @service informationBar!: InformationBarService;
+    transition = slideDown;
+
+    /**
+     * Returns whether or not the loading bar is visible
+     *
+     * @readonly
+     * @type {boolean}
+     */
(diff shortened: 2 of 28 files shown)


pep/app/styles/_variables.scss
```diff
@@ -15,6 +15,7 @@ $progress-spinner-loading-size: 15px;
 $progress-spinner-loading-border-width-lg: 4px;
 $progress-spinner-loading-size-lg: 50px;
 $drag-bar-thickness: 5px;
+$information-bar-zindex: 99999999;
```
billdami, line 18: can you move this to the `//Z-INDICES` group below, and rename it to follow the `$z-*` convention

pep/app/styles/components/_information-bar.scss
```diff
@@ -0,0 +1,12 @@
...
+    position: absolute;
+    top: 0;
+    left: 0;
+    right: 0;
+    background: $color-information-bar-bg;
+    transition: opacity 0.2s;
+    background-clip: padding-box;
+    overflow: hidden;
+    will-change: opacity;
+    z-index: $information-bar-zindex; // kt-aside + kt-header have z-index of 97 & 98 respectively
```
billdami, line 11: whats "kt-aside + kt-header"?

pep/app/utils/user.ts
```diff
@@ -0,0 +1,24 @@
+import { getOwner } from '@ember/application';
+
+export async function onAuthenticated(owner: any) {
+    const currentUserService = getOwner(owner).lookup(`service:current-user`);
```
billdami, line 4: would dry up all these getOwner() calls to a `const owner = getOwner(this)`

pep/app/pods/application/template.hbs
```diff
@@ -14,4 +14,5 @@
 </div>
 <Page::Drawer @openAboutModal={{this.openAboutModal}} />
 <NotificationContainer @position="top" />
-<ModalOutlet />
\ No newline at end of file
+<ModalOutlet />
+<InformationBar class="app-information-bar" @light={{true}} />
```
billdami, line 18: doesnt look like this component has a `@light` arg

pep/app/pods/components/information-bar/component.ts
```diff
@@ -0,0 +1,18 @@
...
+import InformationBarService from 'pep/services/information-bar';
+import { slideDown } from 'pep/utils/animation';
+
+interface InformationBarArgs {}
+
+export default class InformationBar extends Component<InformationBarArgs> {
+    @service informationBar!: InformationBarService;
+    transition = slideDown;
+
+    @action
```
billdami, line 14: missing doc block

billdami (changes requested):

billdami (approved):

PR #262, opened 2021-01-29
Bug/gh253 original pdf

### LINKED ISSUE #253: Handling Missing PDF Originals
It's normal for PDF Originals to be missing, since we don't have (and can't get) some of the article originals.  I'd estimate we don't have around 20% of them.

When there's no original, while it would be nicest for the interface to disable that option, I understand that's not really practical in the current architecture, unless I made an API call to check, or unless you directly went to the S3 repository.   For now, it's ok enabled.

HOWEVER: When the user requests an original, and it's not available, i.e.m, The interface should bring up a dialog with a nice, user-oriented message "The PDF Original you requested is not currently available on PEP-Web."

Instead of switching to the server return error message data as a page:

It's quite startling to the user:

https://user-images.githubusercontent.com/4837554/105874899-234a3e00-5fcb-11eb-94ed-14eed532e4e3.mp4

bakerac4:
Fixes #253

### DIFF
--- a/pep/app/pods/components/search/preview/template.hbs
+++ b/pep/app/pods/components/search/preview/template.hbs
@@ -92,33 +92,7 @@
                                     @action={{this.viewDocument}}
                                     class="mr-2"
                                 />
-                                <DropdownMenu @horizontalPosition="auto-left" as |DD|>
-                                    <DD.Trigger @htmlTag="button" class="btn btn-link px-2">
-                                        <FaIcon @icon="download" />
-                                        {{t "search.preview.download.label"}}
-                                        <FaIcon @icon="angle-down" class="ml-1 text-muted" />
-                                    </DD.Trigger>
-                                    <DD.Content class="dropdown-menu">
-                                        <DropdownAction
-                                            @dd={{DD}}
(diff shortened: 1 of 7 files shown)


pep/app/pods/components/search/preview/template.hbs
```diff
@@ -92,33 +92,7 @@
                                     @action={{this.viewDocument}}
                                     class="mr-2"
                                 />
-                                <DropdownMenu @horizontalPosition="auto-left" as |DD|>
-                                    <DD.Trigger @htmlTag="button" class="btn btn-link px-2">
-                                        <FaIcon @icon="download" />
-                                        {{t "search.preview.download.label"}}
-                                        <FaIcon @icon="angle-down" class="ml-1 text-muted" />
```
billdami, line 99: i think we should make an arg in `Document::Actions::Download` to switch to this "expanded" button  content to use here, as it has the room for the more descriptive text/caret and I think it looked quite good in the original preview layout.

billdami (changes requested):

billdami (approved):

Issue #250, opened 2021-01-22
Breakpoint for wrapping too large?


nrshapiro:
1) Watch the following video shot in inspect mode...note the width at the top and where the columns in the Expert Picks break: At a window size of 1513 pixels, which is pretty large (for many people).

I believe it's way too soon to need to break.  Note that I have "small" selected as my base font size.

2) Also note that the image goes on top...I think the image should go to the bottom.  

https://user-images.githubusercontent.com/4837554/105531996-2e922680-5cb8-11eb-993b-0870b289e23b.mp4

Issue #227, opened 2021-01-11
Top Display Clipped with Extra Large Font


nrshapiro:
The default font size change (R7) is nice.

However, on my display, in Opera, the Extra Large Font size cuts off items (mainly "Back to Search", others perhaps by a smidgen) at the top of the window, and there does not seem to be a way to get them back.  Other items overfill their spaces, e.g., in the search form, but the user can fix that easily by making the left pane bigger.  But there's no way to fix the cutoff at the top that I see.

![image](https://user-images.githubusercontent.com/4837554/104195515-b773ad00-53f0-11eb-820d-afe26ea79294.png)

A minor issue, but should be fixed.

nrshapiro:
@adistasio @bakerac4 @billdami @napafundi 

Note that this occurs on smaller screens with any font, as I've been testing on ipad and phone like devices, etc.  in Opera simulation modes at least

Issue #198, opened 2020-12-22
Two video browse issues


nrshapiro:
1) When you click on items in browse, the left pane is supposed to fill with articles from the same issue.  This works for many of the entries (e.g., IJPVS.001A.0003A), but not for some:
   a)  PEPGRANTVS - does not work, [should show this list](https://stage.pep-web.rocks/search?q=art_id:PEPGRANTVS.*.*)
   b)  PEPTOPAUTHVS - does not work, [should show this list](https://stage.pep-web.rocks/search?q=art_id:PEPTOPAUTHVS.*.*)
   c)  AFCVS doesn't work, [should show this list](https://stage.pep-web.rocks/search?q=art_id:AFCVS.*.*) 
   d) IPSAVS doesn't work, [should show this list](https://stage.pep-web.rocks/search?q=art_id%3AIPSAVS.*.*)
   e) NYPSIVS doesn't work, though there's one article, but it [should show this list](https://stage.pep-web.rocks/search?)q=art_id%3ANYPSIVS.*.*
...there may be more...

bakerac4:
@nrshapiro Im probably missing something, but these seem to be working. 

The steps Im following are 

1. Go to /browse/PEPGRANTVS/volumes/1
2. Click on the first item in the list
3. Check the sidebar for those items. I see all 14 items in the sidebar

nrshapiro:
@bakerac4 @adistasio 

Maybe some are working now, I had filled in the bug report form a few days ago...(Friday maybe), but found the tab with the bug this morning as yet unsubmitted, so I submitted it.  BUT...

I just randomly tried a few and found this/these not showing the others:

PEPTOPAUTHVS doesn't work...e.g.,

   https://stage.pep-web.rocks/read/PEPTOPAUTHVS.001.0003A
   https://stage.pep-web.rocks/read/PEPTOPAUTHVS.001.0015A

This one doesn't work:
https://stage.pep-web.rocks/read/PEPGRANTVS.001.0012A

NYPSI (only one article), but I note there's no banner logo

![image](https://user-images.githubusercontent.com/4837554/102925992-429cfb00-4462-11eb-8416-6cec8d9533a3.png)

ALSO: How do users get to /browse/PEPGRANTVS/volumes/1?  It might be nice to be able to, but now we are listing individual videos rather than publishers, so how does a user get there?  I got there by URL manip, not clicking.

[...]

PR #264, opened 2021-02-02
Bug/gh198 video browse

### LINKED ISSUE #259: Conjoined Words in the Glossary Cloud list
@adistasio @bakerac4 

The glossary infocard cloud often shows conjoined words, which are not in the glossary.

The problem was traced back to a [solr bug](https://github.com/Psychoanalytic-Electronic-Publishing/OpenPubArchive-Content-Server/issues/83).  Not sure we can do anything about it.  Open for now. 

Note that the client does not currently do anything with the document get glossary endpoint returns an error.  Not sure if that's better or providing an error dialog, which requires the user to act.  It might be best if the client showed it in one of those transitory messages you see when you log in, so the error is acknowledged, but it requires no effort on the user to dismiss.

Thoughts?

bakerac4:
Fixes #259 , #250, #227, #198


pep/package.json
```diff
@@ -25,6 +25,7 @@
         "abortcontroller-polyfill": "^1.4.0",
         "aws-serverless-express": "^3.3.6",
         "dotenv": "^8.2.0",
+        "ember-intl": "^5.6.2",
```
billdami, line 28: any reason this had to switch to the `dependencies` block?

billdami (changes requested):

billdami (approved):

PR #268, opened 2021-02-02
Fixing dark mode visiblity and mobile formatting

bakerac4:


pep/app/constants/preferences.ts
```diff
@@ -104,7 +104,7 @@ export const DEFAULT_USER_PREFERENCES: UserPreferences = {
     searchHICEnabled: true,
     searchHICLimit: 1,
     readLater: [],
-    theme: ThemeId.DEFAULT,
+    theme: ThemeId.DARK,
```
billdami, line 107: we are making dark mode the default theme for all users? :notsureif:

bakerac4: Good catch. I had to change that to test the issue and forgot to change it back

billdami (approved):

PR #269, opened 2021-02-03
Adding in session to about dialog

bakerac4:


pep/app/pods/components/modal-dialogs/user/about/component.ts
```diff
@@ -30,4 +34,10 @@ export default class ModalDialogsUserAbout extends Component<ModalDialogsUserAbo
     get status(): string {
         return this.intl.t(this.isStatusOk ? 'about.status.ok' : 'about.status.down');
     }
+
+    get sessionId(): string | undefined {
+        return this.session.isAuthenticated
+            ? this.session.data.authenticated.SessionId
+            : this.session?.getUnauthenticatedSession()?.SessionId;
+    }
```
billdami, line 42: i feel like this might be good logic to have directly in the session service itself, i.e. "give me the 'absolute' session ID regardless of auth state", since it seems reasonable we'd want to use that elsewhere int he app. thoughts?

also why does `session` have `?` existence checking in one instance, but not the other 2?

billdami (changes requested):

billdami (approved):

Issue #273, opened 2021-02-05
Hits in Context not highlighting full match in some cases


nrshapiro:
This is likely to be a Solr issue, but I wanted to record it here in case it's a markup/xml/xslt issue.  And also I wanted to make sure the UI issue is recorded where people would look.

I searched for the phrase "death in the family", in quotes (smartsearch).

What was interesting, was that the hit list within the bibliographical result listings were perfect, but when you went to a document, the phrase was not marked up, just the first word, "death".  **BUT** the entire phrase was marked as a hit for ANY DOCUMENT which had multiple hits.  Only the ones with a single match (and it was matching the phrase), had only the first word, death, highlighted.

Document sample with one hit:

![image](https://user-images.githubusercontent.com/4837554/107064618-b80e2200-67a9-11eb-8853-9f50a805a2bf.png)

Document with multiple hits:

![image](https://user-images.githubusercontent.com/4837554/107064533-9c0a8080-67a9-11eb-8bdb-34e189899276.png)

bakerac4:
@nrshapiro this is not a SOLR issue.

It seems to be an issue when there is a glossary term inside a search hit. You can see the issue with multiple search terms in the document by going here /search/document/PSC.034.0547A?preview=PSC.034.0547A&q="death%20in%20the%20family"

bakerac4:
@nrshapiro is it possible for you to adjust the xml you send back to fix this?
```
<impx type="TERM2" rx="ZBK.069.0001d.YN0013300865870" grpname="DEATH">#@@@death</impx> in the family@@@#```

Thats what I currently get back. Would it be possible for you to wrap the search term around the tag like so?

```#@@@<impx type="TERM2" rx="ZBK.069.0001d.YN0013300865870" grpname="DEATH">death</impx> in the family@@@#```

nrshapiro:
@bakerac4 - There's no easy fix for this.  Remember, the XML is stored in Solr.  Solr is marking hits without having any rules regarding the content or markup of the field.  

Its a Solr problem because Solr doesn't care about the markup and where it's placing the hit boundaries.

[...]

PR #280, opened 2021-02-09
Story/tp66923 wcag

### LINKED ISSUE #275: Indication of selected document missing
Client: 2021.02.04-6035

In read mode, as I click on items in the read list pane, I don't see any indication there of my selection.

image

bakerac4:
Fixes #275 and #273

### DIFF
--- a/pep/app/constants/search.ts
+++ b/pep/app/constants/search.ts
@@ -659,3 +659,9 @@ export const HIT_MARKER_END = '@@@#';
 export const HIT_MARKER_START_OUTPUT_HTML = `<span class='searchhit' data-type='${DocumentLinkTypes.SEARCH_HIT_TEXT}'>`;
 export const HIT_MARKER_END_OUTPUT_HTML = '</span>';
 export const SEARCH_HIT_MARKER_REGEX = new RegExp(`${HIT_MARKER_START}|${HIT_MARKER_END}`, 'g');
+
+// Finds all tags with a #@@@ inside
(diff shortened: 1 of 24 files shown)


pep/app/constants/search.ts
```diff
@@ -659,3 +659,9 @@ export const HIT_MARKER_END = '@@@#';
 export const HIT_MARKER_START_OUTPUT_HTML = `<span class='searchhit' data-type='${DocumentLinkTypes.SEARCH_HIT_TEXT}'>`;
 export const HIT_MARKER_END_OUTPUT_HTML = '</span>';
 export const SEARCH_HIT_MARKER_REGEX = new RegExp(`${HIT_MARKER_START}|${HIT_MARKER_END}`, 'g');
+
+// Finds all tags with a #@@@ inside
+export const POSSIBLE_INVALID_SEARCH_HITS = /<[^>]*>(([^<>/])*?(#@@@))([^<>/])*?<\/[^>]*>/g;
```
billdami, line 664: you should be able to change the `*`'s  in `<[^>]*>` and `<\/[^>]*>` to `+`'s I think, as plus sign matches if theres 1 or more, whereas asterisk matches zero or more, so its a little more strict/safe.

pep/tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2017",
+    "target": "es2020",
```
billdami, line 3: any particular reason you're bumping the compile target up to ES2020? not specifically against it and I know we targeting mostly evergreen browsers in this app, but just want to make sure we don't cause any browser incompatibility issues by going for the bleeding edge in our transpiling.

billdami (changes requested):

billdami (approved):

PR #299, opened 2021-02-25
Story/tp67160 admin

bakerac4:


pep/app/pods/application/adapter.ts
```diff
@@ -82,6 +82,7 @@ export default class Application extends DS.RESTAdapter.extend(FastbootAdapter)
             const xForwardedFor = (fastbootHeaders.get('X-Forwarded-For') as string) ?? '';
             headers['X-Forwarded-For-PEP'] = xForwardedFor.split(',')[0];
         }
+        headers['x-api-authorize'] = 'pr2x%$ddE??233aGYKSE';
```
billdami, line 85: i think this value should definitely be stored/injected  in our .env files, even if for now it doesnt change per environment, just to follow our usual patterns for organization of having all api-key like values in one place (and as something that has a decent chance of changing later, so even if its the same everywhere, having it injected .env at build-time means we dont need to do a code change to do so)

also if this is something thats added for every request, probably should be defined in the initial `const header` where `client-id` is set.

pep/translations/en-us.json
```diff
@@ -61,6 +61,14 @@
...
             "preferences": "Help Preferences",
             "provideFeedback": "Provide Feedback"
+        },
+        "admin": {
+            "common": "Common",
+            "languages": {
+                "enUs": "English",
+                "esEs": "Spanish",
+                "frFr": "French"
+            }
```
billdami, line 71: just labeling the nav item (and page title) "Common" seems a little vague, what about "Common Settings"?

additionally, it might be nice to below the dropdown divider have a dropdown-title (i think thats the bootstrap class) heading that is "Language Settings" or "Content Settings".

pep/app/validations/configuration/common.ts
```diff
@@ -0,0 +1,24 @@
...
+                    code: [validatePresence({ presence: true, ignoreBlank: true })]
+                },
+                whatsNew: {
+                    limit: [validatePresence({ presence: true, ignoreBlank: true })]
+                },
+                mostCited: {
+                    limit: [validatePresence({ presence: true, ignoreBlank: true })]
+                },
+                mostViewed: {
+                    limit: [validatePresence({ presence: true, ignoreBlank: true })]
```
billdami, line 17: should these possibly have a `validateNumber()` validation as well?

pep/app/services/configuration.ts
```diff
@@ -27,7 +24,7 @@ export default class ConfigurationService extends Service {
      * @returns {string}
      */
     get contentConfigName() {
-        return `${CONTENT_CONFIG_NAME}-${this.lang.currentLanguage}`;
+        return `${this.lang.currentLanguage}`;
```
billdami, line 27: shouldn't need the string interpolation here anymore, i dont think.

pep/app/router.js
```diff
@@ -28,6 +28,12 @@ Router.map(function() {
 
     this.route('most-cited');
     this.route('most-viewed');
+    this.route('admin', function() {
+        this.route('common');
+        this.route('en-us');
+        this.route('fr-fr');
+        this.route('es-es');
```
billdami, line 35: as mentioned above, wouldnt need individual routes defined w/the dynamic solution

billdami (changes requested):

pep/app/pods/admin/common/template.hbs
```diff
@@ -0,0 +1,104 @@
...
+                        <div class="col-12">
+                            <validator.input
+                                @errors={{changeset.error.configSettings.global.cards.videoPreview.code.validation}}
+                                @label={{true}}
+                            >
+                                <FlInput
+                                    @value={{changeset.configSettings.global.cards.videoPreview.code}}
+                                    @placeholder={{t "admin.common.global.cards.videoPreview.code"}}
+                                >
+                                    <textarea
```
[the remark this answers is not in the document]
bakerac4: Agreed. Its on the to-do list

pep/app/pods/application/adapter.ts
bakerac4: Yea, this is totally temporary. Will figure out a different solution IF we keep it

billdami: ok, as long as you aware Past-Adam will be held fully accountable for this if we end up leaving it and forgetting about it. 😛

pep/app/constants/configuration.ts
```diff
@@ -100,16 +100,8 @@ export interface ContentConfiguration {
...
-                subheading?: string;
-                body: string;
-            };
-            right: {
-                heading: string;
-                subheading?: string;
-                body: string;
-            };
+            left: string;
+            right: string;
```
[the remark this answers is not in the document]
billdami: @adistasio thoughts, feelings, concerns?

billdami (approved):

PR #300, opened 2021-02-25
Adding in referrer header to IP call

### LINKED ISSUE #293: Referrer Login direct to Client
@adistasio @nrshapiro 
As discussed at this weeks meeting we need to allow for "referrer" logons going direct to the client. I have now changed the pads IP authenticate API so that if it is passed a referrer URL in the header it will try and use that, if the IP address is not registered. If the logon is successful the logon method will be ReferrerURL.

Pads will be looking for a header item called "ReferrerURL-For-PEP".  Please note that referrer is spelt correctly unlike the HTML standard.

bakerac4:
Fixes #293

### DIFF
--- a/pep/app/authenticators/ip.ts
+++ b/pep/app/authenticators/ip.ts
@@ -31,13 +31,12 @@ export default class IpAuthenticator extends CredentialsAuthenticator {
             }
             const headers = this.authenticationHeaders;
             if (this.fastboot.isFastBoot) {
-                const referrer = this.fastboot.request.headers.headers['referer'];
+                const referrer = this.fastboot.request.headers?.get('referer');
 
                 if (referrer) {
                     headers['ReferrerURL-For-PEP'] = referrer;
                 }
             }
-            console.log(headers);
             const response = await this.ajax.request<PepSecureAuthenticatedData>(url, {
                 headers
             });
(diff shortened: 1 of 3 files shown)


pep/app/authenticators/ip.ts
```diff
@@ -29,9 +29,17 @@ export default class IpAuthenticator extends CredentialsAuthenticator {
...
             }
+            const headers = this.authenticationHeaders;
+            if (this.fastboot.isFastBoot) {
+                const referrer = this.fastboot.request.headers.headers['referer'];
 
+                if (referrer) {
+                    headers['ReferrerURL-For-PEP'] = referrer;
+                }
+            }
+            console.log(headers);
```
billdami, line 40: stray `console.log()`, should that be left in?

pep/app/authenticators/ip.ts
```diff
@@ -29,9 +29,17 @@ export default class IpAuthenticator extends CredentialsAuthenticator {
                 const params = serializeQueryParams({ SessionId: sessionData.SessionId });
                 url = `${url}?${params}`;
             }
+            const headers = this.authenticationHeaders;
+            if (this.fastboot.isFastBoot) {
+                const referrer = this.fastboot.request.headers.headers['referer'];
```
billdami, line 34: is `fastboot.request.headers` an actual `Headers` api instance? if so, it might be better to do `this.fastboot.request.headers.get('referer')`

billdami (changes requested):

billdami (approved):

PR #303, opened 2021-03-03
Story/tp67167 admin

bakerac4:


pep/app/pods/components/document/text/component.ts
```diff
@@ -457,14 +462,50 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
+    async setupListeners(element: HTMLElement) {
         this.containerElement = element;
         this.scrollableElement = this.containerElement?.closest('.page-content-inner');
         scheduleOnce('afterRender', this, this.attachTooltips);
         if (this.args.page) {
-            this.scrollToPageOrTarget(this.args.page);
+            await this.scrollToPageOrTarget(this.args.page);
         }
+        Object.values(this.containerElement.querySelectorAll('div.pagebreak')).forEach((item) => {
+            const observer = new IntersectionObserver(
```
billdami, line 480: @bakerac4 not that it'll fix your query param issue, but I just noticed this. Looking the `IntersectionObserver` I _think_ you should be able to create just one instance of it, and in your loop here you only call `observer.observe()` for each element. That way you don't have like hundreds of `IntersectionObserver`'s on really long documents.

pep/tests/integration/helpers/embargo-year-test.ts
```diff
@@ -0,0 +1,19 @@
...
+module('Integration | Helper | embargo-year', function(hooks) {
+    setupRenderingTest(hooks);
+
+    // Replace this with your real tests.
+    test('it renders', async function(assert) {
+        this.set('inputValue', '1234');
+
+        await render(hbs`{{embargo-year inputValue}}`);
+
+        assert.equal(this.element.textContent?.trim(), '1234');
```
billdami, line 17: doesnt look like that would pass based on what the helper is doing, i dont _think_.

pep/app/pods/components/tables/cell/actions/component.ts
```diff
@@ -0,0 +1,11 @@
+import Component from '@glimmer/component';
+
+interface TablesCellActionsArgs {}
```
billdami, line 3: fine for now, but we probably should have a base interface that has `rowValue`, `cellValue`, etc args defined (possibly in our table addon), that we can extend from for any cell components like this.

pep/app/pods/admin/route.ts
```diff
@@ -4,11 +4,24 @@ import Route from '@ember/routing/route';
...
+    /**
+     * Check and see if the user can access this route, and if not forward them to the 403 page
+     *
+     * @memberof Admin
+     */
+    beforeModel(): void {
+        const access = canAccessRoute(this, ['user.viewAdmin']);
+        if (!access) {
+            this.transitionTo('four-oh-three');
+        }
```
billdami, line 21: since this use case is so common, i would probably package this up completely into a helper like `authorizedRoute(this, ['ability', ...])` (and use the router service for the transition) so you can just do:

```ts
beforeModel() {
    authorizedRoute(this, ['ability', ...])
}
```

pep/app/pods/components/document/text/component.ts
```diff
@@ -607,4 +651,13 @@ export default class DocumentText extends Component<DocumentTextArgs> {
...
     }
+
+    /**
+     * Turn off page tracking so even if the user happens to scroll before the component is destroyed, we dont update the page
+     *
+     * @memberof DocumentText
+     */
+    willDestroy(): void {
+        this.pageTracking = false;
+    }
```
billdami, line 662: this is fine, but using a concurrency task would've handled this for you too, w/o the need for the additional `pageTracking` state and this extra boilerplate.

pep/app/pods/components/home/component.ts
```diff
@@ -39,22 +42,42 @@ export default class Home extends Component<HomeArgs> {
...
+    saveExpertPickViewToCookie(index: number): void {
+        this.cookies.write(
+            this.cookieName,
+            JSON.stringify({
+                index,
+                dateViewed: new Date()
+            }),
+            {}
+        );
+    }
```
billdami, line 79: this implementation (using cookies to determine which expert pick to show) seems like it might be at odds with what they really want here, as this means different users (and even the same user on different browsers/devices) will potentially see a different "pick of the day", depending on when they first viewed, right?

@adistasio can correct me if I'm wrong but I assumed they'd want the "pick of the day" to be the same for everyone on any given day? We could actually do this pretty easily, and w/alot less complexity/no cookies, if we just added another config, `home.expertPicksStartDate` where they enter a date value. Then on render of the page you simply get the difference between that date and the current date, and that's your index of which expert pick to show in the list (and if its greater than the list length, you just loop back around).

billdami (changes requested):

pep/app/pods/components/home/component.ts
adistasio: @billdami yes, do the start date

pep/app/pods/components/home/component.ts
```diff
@@ -122,19 +126,28 @@ export default class Home extends Component<HomeArgs> {
...
      */
     @action
-    async loadModel() {
+    async loadModel(): Promise<void> {
         const expertPicks = this.configuration.base.home.expertPicks;
-        const result = await this.store.findRecord('abstract', expertPicks[expertPicks.length - 1].articleId);
+        const specialDate = this.configuration.base.home.expertPicksStartDate;
+
+        const dayDifferential = moment(new Date()).diff(specialDate, 'days');
+        const index = dayDifferential % expertPicks.length;
```
billdami, line 140: like the `expertPick()` getter i _think_ you could bundle up the `index` calculation into its own getter too, and would also remove the need for the extra `@tracked` currentlyShownIndex value, since its purely derived state, doesnt need to be mutable:

```ts
get expertPickIndex() {
    const expertPicks = this.configuration.base.home.expertPicks;
    const specialDate = this.configuration.base.home.expertPicksStartDate;
    const dayDifferential = moment(new Date()).diff(specialDate, 'days');
    return  dayDifferential % expertPicks.length;
}

get expertPick() {
    const expertPicks = this.configuration.base.home.expertPicks;
    return expertPicks[this.expertPickIndex];
}
```

billdami (approved):

PR #309, opened 2021-03-05
feature/tp67412-67255-misc

napafundi:


pep/app/pods/components/page/nav/component.ts
```diff
@@ -132,4 +134,14 @@ export default class PageNav extends Component<PageNavArgs> {
     openFeedbackModal() {
         return this.modal.open('help/feedback', {});
     }
+
+    /**
+     * Set any necessary tracked properties when
+     * a dropdown is opened.
```
bakerac4, line 140: Should probably add a commend to tell future us WHY we had to do this

PR #315, opened 2021-03-08
Adding intro mode text admin configs

bakerac4:


pep/app/constants/regex.ts
```diff
@@ -10,3 +10,4 @@ export const PARENTHESIS_REGEX = /\((.*)\)/;
 export const NON_DIGIT_REGEX = /\D/g;
 export const SEARCH_STRING_REGEX = /\(([^()]+)\)/g;
 export const SEARCH_STRING_TERMS_REGEX = /[^a-zA-Z\s-]+/g;
+export const URL_REGEX = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/;
```
billdami, line 13: ember-changeset-validations should already have `url` format validation built-in:

https://github.com/poteto/ember-changeset-validations#format

billdami (changes requested):

pep/app/constants/regex.ts
bakerac4: They do - however they don't require the https or http which we do

PR #328, opened 2021-03-11
Making expert pick help icon and adding admin config

bakerac4:


pep/translations/en-us.json
```diff
@@ -846,6 +846,9 @@
         "content": {
             "help": {
                 "title": "Help icon text",
+                "home": {
+                    "expertPickOfTheDay": "Expert Pick of the Day"
+                },
```
billdami, line 851: should this be copied to the other translation files?

billdami (changes requested):

billdami (approved):

PR #329, opened 2021-03-11
67125-not-all-preferences-load-when-logged

### LINKED ISSUE #320: Federated Links
@adistasio @nrshapiro 
At present these are opening in a new tab. They should open in the same tab as when the user completes their Federated logon they will be returned to the client.

### LINKED ISSUE #322: About menu item placement
Now that we have a ? menu, it seems like it would make more sense to put "About" there.  And to distinguish it, and have it stand out, we might want to rename it "About PEP-Web" there. 

Marked as high priority, only because if we agree we should do this, it would be good to do it before the betas.

### LINKED ISSUE #325: Secondary Login – not refreshing
- If I login via a referrer or Federated and then do a secondary login with an individual user it does not refresh the user details, or the logon method, therefore I can’t see who I am or log out. 
- If I take the session id and use it in a query string to the client it looks ok.

### LINKED ISSUE #326: Logout – Federated & Referrer
You should be able to log out if you have logged in using one of these methods. 

You could also have the logout button if you could login via IP, this would log you straight back in again, if your browser is still on the same IP, but I am not sure that that would be a problem.

### LINKED ISSUE #327: Logon Method – not showing for individuals
The logon method is showing for primary Group logins, but not for individual logins either primary or secondary.  

This would be useful but not essential, and it may also help with what you show on the front-end.

napafundi:
Fixes #322 
Fixes #320 
Fixes #326 
Fixes #325
Fixes #327


pep/app/utils/user.ts
```diff
@@ -36,7 +38,8 @@ export async function onAuthenticated(owner: Controller | Route | Component): Pr
     currentUserService.setFontSize(currentUserService.fontSize.id);
     await themeService.setup();
     await langService.setup();
-    return configurationService.setup();
+    configurationService.setup();
```
bakerac4, line 41: I think you'll want an `await` here. That being said, I wonder if it might be worth doing a `Promise.all` so these are not occurring synchronously. At the very least for the `themeService`, `langService` and `configurationService` - I don't believe any of those rely on each other.

PR #334, opened 2021-03-12
Feature/tp66923 wcag

napafundi:


pep/app/pods/components/modal-dialogs/help/report-data-error/template.hbs
```diff
@@ -64,49 +65,57 @@
...
+                            <legend class="pl-md-1 d-flex radio-input-legend">
+                                {{t "reportDataError.form.authorOrPublisher"}}
+                            </legend>
+                            <div class="d-flex">
+                                <RadioInput
+                                    @label={{t "common.yes"}}
+                                    @value={{true}}
+                                    @checked={{changeset.isAuthorPublisher}}
+                                    @onChange={{fn this.updateBoolean "isAuthorPublisher" true}}
+                                    @name={{t "reportDataError.form.authorOrPublisherName"}}
```
billdami, line 79: I dont think its a WCAG requirement that we have a translated `name=""` attribute here, as long as the inputs have a correctly linked `<label>` (e.g. with a `for=""` attribute, etc), which I thought these `<RadioInput>`'s do have. 

That being said its not hurting anything to have these here, other than maybe making the translation files bigger and harder to translate (since the translator wont even be able to see these strings in the page)

pep/app/utils/string.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Generates a cleaned up and dasherized
+ * label for accessibility use.
+ *
+ * @export
+ * @param {string} str
+ * @returns {string}
+ */
+export function createAccessibilityName(str?: string) {
```
billdami, line 11: id maybe name this a bit more generically since it could be used for other things, maybe like `cleanAndDasherize()`?

And just note in the doc block that its removing non-word characters specifcallly.

billdami (changes requested):

pep/app/pods/components/modal-dialogs/help/report-data-error/template.hbs
napafundi: From the examples that I looked through for this, most of the time I noticed that each `<input>` had the `name=""` attribute which referenced the topic(?) of what you're making a selection for. But yeah it might not be in the WCAG spec, especially since the `<legend>` element basically does the input-topic labeling for you.

billdami: yeah, thats mainly for traditional server-side, forms, as that `name` is how you access its value in the POST data. As far as I've seen in my research on WCAG stuff, as long as you have the normal visible identifying elements like `legend`, `label` properly hooked up, the `name` attribute is only used as an absolute last resort fall back as far as screenreaders and stuff go. Especially since alot of the time the name value can be somewhat non human-friendly to read.

but yeah like i said im fine withe keeping this in.

billdami (approved):

PR #352, opened 2021-03-16
Fixing auth issue

bakerac4:


pep/package.json
```diff
@@ -43,7 +43,7 @@
         "@gavant/ember-floating-labels": "^1.1.0",
         "@gavant/ember-modals": "^1.0.3",
         "@gavant/ember-pagination": "3.0.0-beta.2",
-        "@gavant/ember-shopify-draggable": "^0.3.3",
+        "@gavant/ember-shopify-draggable": "https://github.com/Gavant/ember-shopify-draggable.git#update-addon",
```
billdami, line 46: do you really need master to use this unpublished branch for the addon? if you think its ready to go can we just publish a beta release, or if you are trying to test stage/prod specific issues, using at least a specific commit hash here instead might be best. 

My worry is if you happen to push to that branch later on, subsequent prod builds will pick up those changes, so its a little volatile.

pep/app/session-stores/application.js
```diff
@@ -1,10 +1,10 @@
...
 
-export const SESSION_COOKIE_NAME = 'pep_session';
+export const SESSION_COOKIE_NAME = 'ember_simple_auth-session';
 
 export default CookieStore.extend({
-    cookieName: SESSION_COOKIE_NAME,
-    cookieExpirationTime: 60 * 60 * 24 * 365, // 1 year
-    sameSite: ENV.cookieSameSite
+    // cookieName: SESSION_COOKIE_NAME,
+    // cookieExpirationTime: 60 * 60 * 24 * 365 // 1 year
```
billdami, line 8: won't using the default expiration here cause the cookie to be deleted when the browser is closed?

https://ember-simple-auth.com/api/classes/CookieStore.html#property_cookieExpirationTime

pep/app/pods/components/modal-dialogs/user/login/component.ts
```diff
@@ -66,6 +66,7 @@ export default class ModalDialogsUserLogin extends Component<ModalDialogsUserLog
             await this.args.options.onAuthenticated?.(response);
             return response;
         } catch (err) {
+            console.log(err);
```
billdami, line 69: do we want this console.log() in prod

pep/app/pods/application/route.ts
```diff
@@ -99,11 +99,12 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
         ) {
             try {
                 await this.session.authenticate('authenticator:ip');
+
                 // We are doing this to work around a strange cookie issue seen from what we think is Ember Simple Auth.
                 // The fastboot session would authenticate through IP, we would write to the cookie and save everything but when the app
                 // loaded back up, we would be logged out and the cookie would mysteriously be removed. We work around this by writing the our own cookie
                 // and then getting that cookie and setting it as the simple auth cookie to force auth state.
-                this.cookies.write('pep_session_fastboot', this.cookies.read('pep_session'), {});
+                this.cookies.write('pep_session_fastboot', this.cookies.read('ember_simple_auth-session'), {});
```
billdami, line 107: should we use the `SESSION_COOKIE_NAME` constant here

pep/app/pods/application/route.ts
```diff
@@ -112,7 +113,7 @@ export default class Application extends PageLayout(Route.extend(ApplicationRout
             // and pass this new session to the user endpoint to make sure we get the correct one. We can't just rely on having the right ID
             // as the update from cookie to session service happens asynchronously
             const newSession = this.cookies.read('pep_session_fastboot');
-            this.cookies.write('pep_session', newSession, {});
+            this.cookies.write('ember_simple_auth-session', newSession, {});
```
billdami, line 116: here as well

billdami (changes requested):

pep/package.json
bakerac4: Whoops,  that actually shouldn't be there. I can fix that

pep/app/session-stores/application.js
bakerac4: Yes. Was it a requirement for it to last a year? Just trying to keep this to the default if possible

billdami: Thats what Alan had me set it to when I had originally asked him. I think the thought was that since we dont get the real session expiration back from the API upon login, we would just make it something super long, to ensure that its always the API that invalidates the user's session.

I feel like for normal user logins, they are gonna kick this back as a bug, if you get logged out whenever the browser is closed, but would check w/ @adistasio to be sure.

billdami (approved):

Issue #311, opened 2021-03-06
Admin Mode - Add Login Dialog Text (but see Issue #312 which changes the login dialog)


nrshapiro:
![image](https://user-images.githubusercontent.com/4837554/110211157-4afeb280-7e63-11eb-8466-6dcad82e4774.png)

The message on the login dialog changes periodically by David and Athena, who is in charge of subscriptions.  I already had a message, this morning to change it, and I'd rather let them change it themselves directly, unless that presents a problem.

It may end up being larger as well.  I assume the dialog can resize dynamically.  As you can see, there's a lot of text in the current site message (not all of which applies, i.e., the second paragraph):

![image](https://user-images.githubusercontent.com/4837554/110211241-b3e62a80-7e63-11eb-84d6-702ded5c3b26.png)

The initial login change request for the text that came this morning:

> This is the Psychoanalytic Electronic Publishing Archive full text database. It is freely available for everyone to search, view tables of contents, and view summaries. However you will need a subscription in order to view full text, follow links, etc. Users with personal access codes can use all personalisation features immediately. Users accessing via a university or group without need to register (free) with PEP to use them.

_and included instructions for which text is not finalized (and neither may be the above):_

> Need a link to subscribe - http://support.pep-web.org/subscribe/ at present and also how to register. (Registration needs ref to privacy rules http://support.pep-web.org/subscribe/pads-and-its-data-protection-policy/ needs updating to cover data we are collecting.

The second part of the above will probably go below the two edit fields, so we need a template of sorts where the text can specify where to place those, or two fields (above/below).

If this presents a problem or too much work prior to Beta 1, let us know and we can try to work it out within the strings settings.  I guess we could alternatively work directly with the PEP-Web-User-Interface git branches and do the string settings there.  But there may not be two different settings currently as is anyway (above/below fields).

nrshapiro:
See #312 which changes the login dialog to a infocard and could have implications for the way it's edited in admin

nrshapiro:
@adistasio 

This is listed now, in case the dialog is to persist in addition to the infocard, and to keep note of the document text and link changes that had been requested.

adistasio:
You can edit the dialog and side login via admin. Closing.

PR #354, opened 2021-03-17
feature/tp67520-login-widget

### LINKED ISSUE #312: "In your face" request to login.
The short login message at the top is supposed to be an in-your-face prompt to encourage logins, something PEP is very keen on.  However, I am not sure when the top login box comes up, since it mainly appears for me when I log out, or if I do a cookie clean, and then go through the intro mode dialogs.  Once you dismiss it, there's nothing that encourages you to log in.

But David is very keen on having the login prompt "in the user's face" until they log in.  So I came up with this solution.

We can instead use an Infocard for the login, rather than a popup dialog.  That way, it can stay loaded while the user is not logged in, as a reminder.  It may even be used in addition to the top prompt that comes up.

This Sign In infocard will always be visible at the top of related info UNLESS you are logged in.  That way, the incentive and information to log in is always there, until such time as you are logged in.  Once you log in, it will hide itself until you log out.  At that point, it will reappear.  It will not, of course, be an infocard that the user can turn off in preferences.  In fact, I think it should not be "foldable" like the other cards.  But like the other cards, Admin should be able to determine where it appears (left or right).

Here's a mockup which David has approved.  The actual text needs to be configurable in Admin:

image

In addition, please add a message to the Preferences dialog warning the user that unless they are logged in, preferences will not be saved.

(Beta 1 if possilble...but I've listed this as Beta 2)

napafundi:
Fixes #312


pep/app/pods/components/page/sidebar/widgets/sign-in/template.hbs
```diff
@@ -0,0 +1,63 @@
+<div class="bg-primary border-divider-b" {{did-insert this.onElementInsert}}>
+    <div class="p-3">
+        <div class="d-flex justify-content-between align-items-center mb-3">
+            <h5 class="mb-0">
+                {{t "relatedInfo.widgets.signIn.title"}}
+            </h5>
+        </div>
+        <ContentWithPlaceholder @placeholderComponent="loading/widget" @isLoading={{this.loadResults.isRunning}}>
```
bakerac4, line 8: @napafundi ideally we would create a component just for the login form so that all this logic is not duplicated

napafundi: @bakerac4 Yeah I agree. I think it might be better to wait to extract that into a component when this is done https://github.com/Psychoanalytic-Electronic-Publishing/PEP-Web-User-Interface/issues/311 but if you think it should be done first I can knock it out

pep/app/pods/components/page/sidebar/widgets/sign-in/component.ts
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * Load the widget results on render
+     *
+     */
+    @action
+    @dontRunInFastboot
+    onElementInsert() {
+        later(() => {
+            taskFor(this.loadResults).perform();
+        }, 1000);
```
billdami, line 125: shouldnt have this delay for this card, I don't think. these delays were just a workaround for some other weird race condition with the main PEP API.

PR #361, opened 2021-03-18
Bug/gh333 preview window

### LINKED ISSUE #333: Preview Window forgets and disappears
I was using the preview window in browse, and I thought I better recheck its operations.

1) set it to not resize each time, but make it a good size.
2) It works nicely
3) Go to an article (read)
4) Go to the Table of Volumes
5) Pick a volume
6) See the TOC
7) Click on articles in the TOC.  There's no preview window anymore.
8) You can get it back though by dragging from the bottom.  

BUG: But it should have kept the settings/size I set.

https://user-images.githubusercontent.com/4837554/110989889-91ab4b80-8340-11eb-8858-7dbc4a41cb3a.mp4

bakerac4:
Fixes #333


bakerac4:
@billdami - I also remembered why we we used pep-session instead of just session as the filename. Its due to the TS-JS conflicts for the same file name. So I switched it back 🙃

billdami (approved):

PR #363, opened 2021-03-19
Fixing browser data

bakerac4:


pep/app/pods/components/page/nav/component.ts
```diff
@@ -28,6 +29,11 @@ export default class PageNav extends Component<PageNavArgs> {
 
     supportUrl = SUPPORT_URL;
     facebookUrl = PEP_FACEBOOK_URL;
+    environment = ENV.environment;
+
+    get isDevelopment(): boolean {
+        return this.environment === 'development';
+    }
```
billdami, line 36: I think we should at least create a quick helper for this, `{{#if (isDevelopment)}}`, to limit the amount of existing code this logic (that kinda, sorta shouldn't be there) touches. And to make it easier to refactor out into named blocks or something later.

billdami (changes requested):

pep/app/pods/components/page/nav/component.ts
bakerac4: If I end up doing this again I would agree with that. But Im not sure there is a need right now as its only in one place?

billdami: I would say its still worthwhile to do, even with it only being used in one spot right now. Mainly to keep this non-app related logic as self-contained as possible.

billdami (approved):

PR #368, opened 2021-03-22
Fixing missing formatting links

### LINKED ISSUE #308: Missing Formatting and Links: Article Keywords
Articles look good in general, but if you compare to our existing system (or Taylor and Francis, since we have some new data like Keywords showing), there are some refinements needed:

1.  The keyword set is not labeled, nor linked.  At a minimum, it should look have the label Keywords, like the T&F site
2.  The keywords are marked <impx> in the file, e.g., 
     - <artkwds><impx type="KEYWORD">autistic spectrum disorder</impx><impx type="KEYWORD">malignant symbiosis</impx><impx type="KEYWORD">narcissistic intrusive-identification</impx><impx type="KEYWORD">osmotic/diffuse identification</impx><impx type="KEYWORD">infantile vulnerability</impx><impx type="KEYWORD">intimacy</impx></artkwds>
3.  The font should also be a separating factor of these.  Perhaps smaller than the main font.

Per the intent of impx, these are implicit links made into explicit links, similar to other impx data.  In the case of keyword implicit links, you would simply do a full-text search based on the clicked keyword phrase, e.g.,.  a fulltext search for "autistic spectrum disorder" per the first item in the list.

image

Taylor and Francis Keyword display:
   
image

bakerac4:
Fixes #308

### DIFF
--- a/pep/app/pods/components/document/read/component.ts
+++ b/pep/app/pods/components/document/read/component.ts
@@ -52,13 +52,13 @@ export default class DocumentRead extends Component<DocumentReadArgs> {
      * @memberof ReadDocument
      */
     @action
-    viewSearch(searchTerms: string) {
+    viewSearch(search: object) {
         // TODO improve this typing
         clearSearch(this as any, this.configuration, this.currentUser);
         this.router.transitionTo('search', {
             queryParams: {
                 ...this.configuration.defaultSearchParams,
-                searchTerms
+                ...search
             }
         });
     }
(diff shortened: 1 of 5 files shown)


pep/app/pods/components/document/read/component.ts
```diff
@@ -52,13 +52,13 @@ export default class DocumentRead extends Component<DocumentReadArgs> {
      * @memberof ReadDocument
      */
     @action
-    viewSearch(searchTerms: string) {
+    viewSearch(search: object) {
```
napafundi, line 55: Any chance we could Type this object? I assume not since you probably would've thought of that 😛

napafundi (approved):

PR #369, opened 2021-03-22
Adding german and italian

### LINKED ISSUE #339: List of App Languages we will support
I talked to DT re about the list of languages we will need to support besides English, in phases:

First tier for beta 1 (already done):
- Spanish
- French

**For beta 2:**
- German 
- Italian

Eventually, maybe:
- Chinese (Mandarin?)
- Russian

This list is per DT based on Psychoanalytic speakers/authors.  For our reference, I wanted to see the most popular spoken languages and found this:

image

bakerac4:
Fixes #339


pep/app/constants/cookies.ts
```diff
@@ -1,5 +1,5 @@
 export const HIDE_TOUR_COOKIE_NAME = 'showTourCookie';
-export const SESSION_COOKIE_NAME = 'ember_simple_auth-session';
+export const SESSION_COOKIE_NAME = 'ember_simple_auth-sessions';
```
billdami, line 2: 🤔

billdami (approved):

billdami:
@bakerac4 you have conflicts in this PR

pep/app/constants/cookies.ts
bakerac4: @billdami I was seeing that really annoying dev only cookie bug.

billdami: ugh 🤦  not that there is a better solution to this, but I hate that we have to do this, especially now that we have real users in the app, since it'll cause dropped sessions and whatnot when this is pushed to prod.

Issue #371, opened 2021-03-22
Problems in the Update Notifications dialog


nrshapiro:
This video illustrates several problem in the Update Notifications dialog: Client: 2021.03.19-6893

1) The email label overlays what the user types for email, making it look bad and difficult to go back and reread
2) It was telling me "No" by flashing the Update Notifications button, but not giving me a clue why (it may be my address is already in the database, but that shouldn't matter, perhaps I'm changing my interests?  And if it's in there, shouldn't it show me what my subscriptions were?  And it should use words to explain what's wrong.

Video:

https://user-images.githubusercontent.com/4837554/112030876-16505380-8b11-11eb-911b-e272285d8d66.mp4

adistasio:
@bakerac4 For this dialog we will remove the email address field. There are then 2 fields SendJournalAlerts, SendVideoAlerts on the save user endpoint in pads. Can only access if logged in as an individual user (or admin). IE no group or unauthenticated users.

PR #378, opened 2021-03-24
Story/gh371 notifications dialog

bakerac4:
Fixes front end for #371

### DIFF
--- a/pep/app/pods/components/modal-dialogs/whats-new/subscription/template.hbs
+++ b/pep/app/pods/components/modal-dialogs/whats-new/subscription/template.hbs
@@ -1,46 +1,33 @@
 <ModalDialog @onClose={{@onClose}} @size="md" @clickOutsideToClose={{false}} as |Modal|>
     <Modal.header @title={{t "relatedInfo.widgets.whatsNew.title"}} />
-    <FormValidator @changeset={{@options.changeset}} @submit={{this.update}} as |changeset validator|>
-        <Modal.body>
-            <p class="text-muted">
-                {{t "relatedInfo.widgets.whatsNew.signUp"}}
-            </p>
-            <Alert
-                @isShown={{bool this.loginError}}
-                @message={{t "login.error"}}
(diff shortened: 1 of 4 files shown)


pep/app/pods/components/modal-dialogs/whats-new/subscription/template.hbs
```diff
@@ -1,46 +1,33 @@
...
-                @action={{validator.submit}}
-            />
-        </Modal.footer>
-    </FormValidator>
+    <Modal.body>
+        <p class="text-muted">
+            {{t "relatedInfo.widgets.whatsNew.signUp"}}
+        </p>
+        <Alert
+            @isShown={{bool this.loginError}}
```
billdami, line 8: do we want to show this alert here if there is a login error? looks like maybe unrelated code.

pep/app/pods/user/model.ts
```diff
@@ -24,19 +26,60 @@ interface ActiveSubscription {
...
+    subscriptionEndDate!: Date;
+    @alwaysSend
+    @attr('string')
+    userFullName!: string;
+    @alwaysSend
+    @attr('string')
+    userName!: string;
+    @alwaysSend
+    @attr('string')
+    userType!: UserType;
```
billdami, line 82: this is alot harder to scan down the list to read model attributes now, imo. if we dont want to keep the single line decorator format here, e.g. `@alwaysSend @attr('string') userName!: string;`, i think we should at least have an empty line between each attribute when using newline-separated decorator format.

billdami (changes requested):

pep/app/pods/user/model.ts
bakerac4: This is how `prettier` automatically formats it. I can add the empty line though

billdami: not very pretty, prettier. 😤

billdami (approved):

PR #385, opened 2021-03-26
Work on keyboard shortcuts

bakerac4:


pep/app/pods/application/controller.ts
```diff
@@ -43,6 +48,85 @@ export default class Application extends Controller {
     @tracked viewedPeriod: ViewPeriod = ViewPeriod.PAST_WEEK;
     @tracked searchTerms: SearchTermValue[] = [];
 
+    @tracked shortcuts: KeyboardShortcut[] = [
```
billdami, line 51: does this need to be a `@tracked` property? looks like its never mutated

pep/app/pods/application/controller.ts
```diff
@@ -43,6 +48,85 @@ export default class Application extends Controller {
...
+                }
+            }
+        },
+        {
+            keys: ESCAPE,
+            shortcut: () => {
+                const active = document.activeElement;
+                const enteringText =
+                    active instanceof HTMLElement &&
+                    (active.isContentEditable || active.tagName === 'INPUT' || active.tagName === 'TEXTAREA');
```
billdami, line 119: this should include `SELECT` since entering text on a select dropdown is valid (it causes the first matching option to be selected), could also clean this up to: 
```ts
['INPUT', 'TEXTAREA', 'SELECT'].includes(active.tagName)
```

pep/app/pods/components/keyboard-shortcuts/component.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    willDestroy() {
+        this.destroyMethod?.();
+    }
+
+    eventHandler(event: KeyboardEvent, callback: (event: KeyboardEvent) => void): void {
+        const active = document.activeElement;
+        const enteringText =
+            active instanceof HTMLElement &&
+            (active.isContentEditable || active.tagName === 'INPUT' || active.tagName === 'TEXTAREA');
```
billdami, line 43: here as well, add `SELECT` to the blacklist, `['INPUT', 'TEXTAREA', 'SELECT'].includes(active.tagName)`

pep/app/pods/components/keyboard-shortcuts/template.hbs
```diff
@@ -0,0 +1 @@
+<div class="shortcut-element" ...attributes {{did-insert this.setupShortcuts}}></div>
```
billdami, line 1: this feels off/non-sematic to me, to have an empty div just to setup keyboard shortcut bindings. I think this is exactly where you would want to create a custom modifier, and then use that on the root element of the applicable templates, or w/e part of it makes sense. e.g.

```hbs
<div class="application-container" {{register-shortcuts this.shortcuts}}>
  ...
</div>
```

```hbs
<nav class="navbar navbar-expand fixed-top navbar-dark bg-gradient-primary" {{register-shortcuts this.shortcuts}} ...attributes>
   ...
</nav>
```
etc.

billdami (changes requested):

pep/app/modifiers/register-keyboard-shortcuts.ts
```diff
@@ -0,0 +1,41 @@
...
+        allowInInput?: boolean;
+    };
+}
+
+interface KeyboardShortcutsArgs {
+    target?: HTMLElement;
+    shortcuts?: KeyboardShortcut[];
+}
+
+export default modifier((_element, _args, named: KeyboardShortcutsArgs) => {
```
napafundi, line 26: Probably should add have a docstring here

pep/app/pods/components/keyboard-shortcuts/component.ts
```diff
@@ -0,0 +1,46 @@
...
+    };
+}
+interface KeyboardShortcutsArgs {
+    target: HTMLElement;
+    shortcuts: KeyboardShortcut[];
+}
+
+export default class KeyboardShortcuts extends Component<KeyboardShortcutsArgs> {
+    destroyMethod?: () => void;
+
```
napafundi, line 20: Docstrings on these methods too

napafundi (changes requested):

PR #416, opened 2021-04-07
Fixing syntax

### LINKED ISSUE #410: Search Issue - Author Field and Facet Query Syntax
Two problems in Author field searches.  I don't think it affects other fields, but I'll have to think more about that.

**Problem 1:**
If a user submits the following search in the interface 

"Tuckett Fonagy"

like the other fields, this is interpreted as an AND by the server. I initially thought this was fine, but depending on the author's name, in many actual use cases it causes problems: for example, when you enter `Cooper, S` in the author field.  It produces 72 results, because it's just looking for Cooper AND S in the author field.  But if you enter the author's name in quotes: `"Cooper, S"` you get 69 matches since in that case Cooper's first initial must be S.  That's much more in line with what users expect.  Especially if they use the word wheel, which definitely can illustrate this problem.

**Fix:**: For the author field, for which the help text doesn't say you can use phrases or booleans, surround the name in quotes (double, not single).  

**However**:  while they can then still use multiple fields to do Ands, then there's no way to do ORs.  So perhaps only surround the field with quotes if there's no boolean connecters?  That would still leave some anomalies though.  Otherwise, if you were to try to quote individual names in a boolean, you would need to parse it and independently quote the names.  We can discuss further if you need. It's probably fine, at least in the short term, to allow the or with some anomalies still in place.

**Problem 2**

After entering the author query, suppose the user selects two facets.

image

The author field submitted to the server is:
```
Tuckett Fonagy AND fonagy, peter OR tuckett, david
```
This returns all tuckett papers, even those without Fonagy.  The core of this problem is the logic in the client...

1) The fields on the top are the query and these are AND'd with the facet fields
2) The facet fields are OR'd together
3) But without parenthetical grouping, the OR is independent.  The submitted field works correctly if it's:

```
     Tuckett Fonagy AND (fonagy, peter OR tuckett, david)
            ^                             ^
         fielded query          facets grouped in parentheses
```

If you do that, then I may also find a query if they enter a boolean OR in the authors field:

(Tuckett OR Fonagy) AND (fonagy, peter OR tuckett, david)

produces 140 matches

vs

Tuckett OR Fonagy AND (fonagy, peter OR tuckett, david)

produces 200 matches.

Note I didn't use quotes as they should be submitted per fix one for first/last name combinations.

The server can't see in these cases which was entered in the author field vs the facet fields.

These two queries produce different results.  

**Fix:** You should group the facet fields in parens.  Like the above, we will need to parenthesize the fielded query portion if ORs are allowed.

bakerac4:
Closes #410

### DIFF
--- a/pep/app/constants/regex.ts
+++ b/pep/app/constants/regex.ts
@@ -11,4 +11,5 @@ export const NON_DIGIT_REGEX = /\D/g;
 export const SEARCH_STRING_REGEX = /\(([^()]+)\)/g;
 export const SEARCH_STRING_TERMS_REGEX = /[^a-zA-Z\s-]+/g;
 export const URL_REGEX = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/;
-export const MEDIA_FILE_EXTENSION = /\.[^/.]+$/;
+export const MEDIA_FILE_EXTENSION_REGEX = /\.[^/.]+$/;
+export const BOOLEAN_OPERATOR_REGEX = /\sOR\s|\sAND\s/;
--- a/pep/app/constants/regex.ts
+++ b/pep/app/constants/regex.ts
@@ -12,4 +12,4 @@ export const SEARCH_STRING_REGEX = /\(([^()]+)\)/g;
 export const SEARCH_STRING_TERMS_REGEX = /[^a-zA-Z\s-]+/g;
 export const URL_REGEX = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/;
 export const MEDIA_FILE_EXTENSION_REGEX = /\.[^/.]+$/;
-export const BOOLEAN_OPERATOR_REGEX = /\sOR\s|\sAND\s/;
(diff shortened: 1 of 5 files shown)


pep/app/constants/regex.ts
```diff
@@ -11,4 +11,5 @@ export const NON_DIGIT_REGEX = /\D/g;
 export const SEARCH_STRING_REGEX = /\(([^()]+)\)/g;
 export const SEARCH_STRING_TERMS_REGEX = /[^a-zA-Z\s-]+/g;
 export const URL_REGEX = /https?:\/\/(www\.)?[-a-zA-Z0-9@:%._\+~#=]{1,256}\.[a-zA-Z0-9()]{1,6}\b([-a-zA-Z0-9()@:%_\+.~#?&//=]*)/;
-export const MEDIA_FILE_EXTENSION = /\.[^/.]+$/;
+export const MEDIA_FILE_EXTENSION_REGEX = /\.[^/.]+$/;
+export const BOOLEAN_OPERATOR_REGEX = /\sOR\s|\sAND\s/;
```
billdami, line 15: per his comment doesnt this also need to allow lowercase and `||` / `&&`?

```suggestion
export const BOOLEAN_OPERATOR_REGEX = /\sOR\s|\sAND\s|\s\|\|\s|\s\&\&\s/i;
```

billdami (changes requested):

pep/app/constants/regex.ts
bakerac4: Yes, seems like he changed his comment after I read it 😭

billdami (approved):

PR #444, opened 2021-04-20
Generalized modal functionality

bakerac4:


pep/app/pods/admin/language/controller.ts
```diff
@@ -229,9 +325,60 @@ export default class AdminLanguage extends Controller {
...
+     * @private
+     * @template T
+     * @param {Paths<LanguageModel>} path
+     * @param {T} item
+     * @memberof AdminLanguage
+     */
+    private createConfigItem<T>(path: Paths<LanguageModel>, item: T) {
+        const items = this.changeset?.get(path) as T[];
+        items.push(item);
+        this.changeset?.set(path, [...items]);
```
billdami, line 343:
```suggestion
        this.changeset?.set(path, [...items, item]);
```

pep/app/pods/admin/language/controller.ts
```diff
@@ -229,9 +325,60 @@ export default class AdminLanguage extends Controller {
...
+     * @template T
+     * @param {Paths<LanguageModel>} path
+     * @param {T} itemToDelete
+     * @param {keyof T} propertyToCompare
+     * @memberof AdminLanguage
+     */
+    private deleteConfigItem<T>(path: Paths<LanguageModel>, itemToDelete: T, propertyToCompare: keyof T) {
+        const items = this.changeset?.get(path) as T[];
+        const filteredItems = items.filter((item) => item[propertyToCompare] !== itemToDelete[propertyToCompare]);
+        this.changeset?.set(path, [...filteredItems]);
```
billdami, line 381:
```suggestion
        this.changeset?.set(path, filteredItems);
```
`filteredItems` is already a new array

billdami (changes requested):

pep/app/pods/admin/language/controller.ts
bakerac4: Filter does create a new array, but the objects referenced in the original are also referenced by the new array. In this case that might be fine, but I think as a pattern it might just be safer to use the spread operator

billdami: the array spread operator will never affect the references to the array's elements, in both cases they'll still reference the original objects, the spread will just create a new array with those same objects, not create copies of each object too. e.g.
```ts
const a = {};
const b = {};
const arr1 = [a, b];
const arr2 = [...arr1];
console.log(arr1[0] === arr2[0]); // -> true
```

If you needed a true deep clone, with new objects for all the elements, you'd probably want to use something like lodash's `cloneDeep`, or you'd have to use an object spread on each element too:
```ts
const arr2 = arr1.map(item => ({ ...item }));
console.log(arr1[0] === arr2[0]); // -> false
```

billdami (approved):

PR #454, opened 2021-04-22
Fixing refresh of page with article list scroll

bakerac4:


pep/app/pods/components/document/read/sidebar/template.hbs
```diff
@@ -113,16 +113,20 @@
...
                             @items={{@paginator.models}}
                             @tagName="ul"
+                            @key="id"
                             @containerSelector=".sidebar-content"
-                            @class="list-unstyled"
                             @estimateHeight={{if this.showHitsInContext 100 20}}
                             @renderAll={{this.fastboot.isFastBoot}}
-                            @lastReached={{@paginator.loadMoreModels}} as |result|
+                            @lastReached={{@paginator.loadMoreModels}}
+                            @idForFirstItem={{@selectedDocument.id}}
```
billdami, line 128: @bakerac4 how does this handle when the initial page(s) of results fetched on load do not include the selected document, e.g. if its like 10 page down or something? Or was this previously handled somewhere else/or are not handling it yet?

billdami (approved):

pep/app/pods/components/document/read/sidebar/template.hbs
bakerac4: Thats handled somewhere else already

PR #471, opened 2021-04-29
Bug/gh462 blank session

### LINKED ISSUE #462: Blank Session
I see this issue is still outstanding. I ran across it again this morning. I went on to pep-web.rocks production and it picked up a session I had been using yesterday.  It appeared as though I was logged on and I could read FULL text documents and make preference changes as normal but when I looked at my user it was blank:
Graphical user interface, text, application

image001
 
Any call to PaDS is returning an error as the session has expired.
 
When I look at the logs when trying to read a document I can see the client asking OPAS for permission:
                http://api.pep-web.rocks/v2/Documents/Document/IJP.060.0311A/?return_format=XML
and I can see OPAS asking PaDS for permission.  Because the session has expired PaDS returns a 401 and “Session has not been authenticated”, but I can still read the document.
 
I have changed the session time out on Stage to 20 minutes so that you can see this problem for yourselves.

bakerac4:
Helps fix #462


pep/tests/unit/authenticators/credentials-authenticator/service-test.ts
```diff
@@ -0,0 +1,89 @@
...
+        const session = this.owner.lookup('service:pep-session');
+
+        const expirationTime = authenticator._absolutizeExpirationTime(expiresIn);
+        const sessionData = { expiresAt: expirationTime, SessionId: '123', IsValidLogon: true };
+        session.setUnauthenticatedSession(sessionData);
+
+        let unauthenticatedSession = session.getUnauthenticatedSession();
+        assert.equal(unauthenticatedSession.SessionId, '123');
+
+        await authenticator.authenticate('Test6', 'Zedra001');
```
billdami, line 74: is it ok for valid site creds, even though its a test account, to be stored in the repo? especially since this a public repo.

cc @adistasio

billdami: @bakerac4 fwiw, i still think these tests are worthwhile, but ideally we would mock the login endpoint with the response you're expecting, and that way you don't have to give it real creds. And also has the added benefit of being alot faster, and not being dependent on the real API being up/functioning, making this more of a true pure unit test (since you're not also testing the api).

but I realize we are under time constraints here, so maybe thats something for the to-do list.

PR #478, opened 2021-04-29
Changing label and sidebar size

bakerac4:


pep/app/constants/dimensions.ts
```diff
@@ -1,6 +1,6 @@
 export const NAV_BAR_HEIGHT = 58;
 export const FOOT_BAR_HEIGHT = 40;
-export const SIDEBAR_WIDTH = 300;
+export const SIDEBAR_WIDTH = 315;
```
billdami, line 3: pretty sure this needs to stay at 300px as the default size, at least for mobile, so that the sidebar toggle/nub is still visible on puny 320px wide phones, otherwise it'll be pushed off screen and they won't be able to close the sidebar. 

Screen Shot 2021-04-30 at 10 08 39 AM

billdami (changes requested):

billdami (approved):

PR #485, opened 2021-05-04
Story/gh477 user context refresh

### LINKED ISSUE #477: User Context and Refresh
Bringing this over from a mention in another bug report, as a general issue.  What I'm writing I just discussed with Alan and it was his point of view as well (and he can chime in if I misrepresent him).

Note that when I do a simple refresh, and I have favorites open, the favorites closes.  

1) Unless the user clears cookies, they should not lose their context when refreshing the page.  
2) That includes when the site times me out and logs me out.  It should offer me an option (even a popup is acceptable there I think) saying something along the lines of:  You have been timed out due to inactivity.  To continue your session, please login" (and that can even be a form of the regular login box letting them do it right there.  One stop shopping.  
3) In contrast, if the user Logs out manually, then you can go back to the home page as it's configured in the admin by default, and then load the user's preferences when they log in

bakerac4:
Fixes #477

### DIFF
--- a/pep/app/pods/components/information-bar/bars/relogin/component.ts
+++ b/pep/app/pods/components/information-bar/bars/relogin/component.ts
@@ -1,0 +1,73 @@
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+import Component from '@glint/environment-ember-loose/glimmer-component';
+import NotificationService from 'ember-cli-notifications/services/notifications';
+import IntlService from 'ember-intl/services/intl';
+
+import AuthService from 'pep/services/auth';
+import CurrentUserService from 'pep/services/current-user';
+import PepSessionService from 'pep/services/pep-session';
+import { BaseGlimmerSignature } from 'pep/utils/types';
+
+interface InformationBarBarsReloginArgs {
+    close: () => void;
+}
+export default class InformationBarBarsRelogin extends Component<BaseGlimmerSignature<InformationBarBarsReloginArgs>> {
+    @service auth!: AuthService;
+    @service currentUser!: CurrentUserService;
+    @service notifications!: NotificationService;
+    @service intl!: IntlService;
+    @service('pep-session') session!: PepSessionService;
+
+    /**
+     * Sets up an auth succeeded event listener to close the information bar
+     * @param {unknown} owner
+     * @param {SearchFormArgs} args
+     */
+    constructor(owner: unknown, args: InformationBarBarsReloginArgs) {
+        super(owner, args);
+        this.session.on('authenticationAndSetupSucceeded', this.onAuthenticationSucceeded);
+    }
+
+    /**
+     * Removes the auth succeeded event listener on component destroy
+     */
+    willDestroy() {
+        super.willDestroy();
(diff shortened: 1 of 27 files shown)


pep/app/pods/components/information-bar/bars/relogin/component.ts
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @memberof InformationBarBarsSettingsAuth
+     */
+    @action
+    login(): void {
+        this.args.close();
+        this.auth.openLoginModal(true);
+    }
+
+    @action
```
billdami, line 62: missing doc block

pep/app/services/current-user.ts
```diff
@@ -413,4 +414,21 @@ export default class CurrentUserService extends Service {
     updateTextJustification(textJustification: TextJustificationId) {
         return this.updatePrefs({ [PreferenceKey.TEXT_JUSTIFICATION]: textJustification });
     }
+
+    modifyWidgetConfigurations(widgetConfigurations: WidgetConfiguration[]) {
+        const configs = this.preferences?.widgetConfigurations;
+        let updatedConfigs = configs;
+        widgetConfigurations.forEach((configuration) => {
+            const configAlreadyExists = configs?.find((item) => item.widget === configuration.widget) ?? false;
```
billdami, line 422:
```suggestion
            const configAlreadyExists = configs?.some((item) => item.widget === configuration.widget);
```
`array.some()` returns a boolean, which it looks like is a better match for what you're doing here.

pep/translations/en-us.json
```diff
@@ -816,6 +816,12 @@
             "buttons": {
                 "signIn": "Sign in"
             }
+        },
+        "relogin": {
+            "title": "You have been logged out. To continue your session, please sign in",
```
billdami, line 821:
```suggestion
            "title": "You have been logged out. To continue your session, please sign in.",
```

billdami (changes requested):

pep/app/services/current-user.ts
```diff
@@ -263,59 +264,79 @@ export default class CurrentUserService extends Service {
...
-                    sameSite: ENV.cookieSameSite,
-                    maxAge: MAX_AGE,
-                    secure: ENV.cookieSecure
-                });
+    /**
+     * Private update prefs method - not to be used by other classes
+     *
+     * @param {PreferenceChangeset} prefValues
+     * @return {*}  {(Promise<Result<UserPreferences | undefined, UserPreferenceError>>)}
+     * @memberof CurrentUserService
```
billdami, line 291:
```suggestion
     * @memberof CurrentUserService
     * @private
```
JSDoc has a `@private` tag it looks like https://jsdoc.app/tags-private.html

billdami (changes requested):

pep/app/services/current-user.ts
bakerac4: Yea Im not concerned about that for now. It looks like truly private methods are coming in TS 4.3 in which case that will probably be auto generated when I use it.

PR #501, opened 2021-05-13
Story/tp68022 allow markup

bakerac4:


pep/app/styles/components/_ember-tooltip.scss
```diff
@@ -0,0 +1,5 @@
+.ember-popover {
+    &-inner {
+        max-width: 250px;
```
billdami, line 3: use `rem`s here if we can

billdami (changes requested):

pep/app/styles/components/_ember-tooltip.scss
bakerac4: I don't think REM's are the correct way to go here. A) the max width is less than a phone width and B) it shouldn't go wider than 250 px because it just looks weird

billdami: id argue `rem` is still applicable in that case, as on desktop you'd want the content container slightly wider/narrower depending on the base size of the text/content, if it is not the default base size. And on mobile ideally you'd have a mobile-breakpoint only style that restricts that max width based on `%` of `vw` to prevent it from going too wide. 

that being said, not a big enough issue to hold this up. 🚀

billdami (approved):

pep/app/styles/components/_ember-tooltip.scss
bakerac4: That and we don't actually display them like a popover/tooltipi in mobile

Issue #503, opened 2021-05-14
In the new html tooltips, is there a way to show a fas icon?


nrshapiro:
The new html tooltips work great.  

I've worked through most of them, but on Related Documents, there's the text {falcon} embedded in the help text for the Related Articles Infocard.  I don't know if that ever showed the right icon...is there a way to show it now that we have some HTML abilities?

I tried the `<i class="fas fa-layer-group"></i>
`
not sure that's the right icon even., but it doesn't work anyway, I presume because I can't add the header include.

I searched for font-awesome ember usage, and saw this as a possible format which was similar to what was there, but it doesn't work here either.

`{{fa-icon "coffee"}}`

nrshapiro:
Also, not sure if it will cause other problems, but can we use a black backgound with white text like before?

bakerac4:
@nrshapiro I can make it a black background if you would like. You _technicallly_ can use FA icons in those tooltips but its convoluted. The newer versions of font awesome use SVG icons by default. Thats what we use, so when its generated it looks something like 

```<svg viewBox="0 0 576 512" xmlns="http://www.w3.org/2000/svg" role="img" focusable="false" aria-hidden="true" data-icon="home" data-prefix="fal" id="ember8" class="svg-inline--fa fa-home fa-w-18 fa-lg ember-view"><path fill="currentColor" d="M541 229.16l-61-49.83v-77.4a6 6 0 0 0-6-6h-20a6 6 0 0 0-6 6v51.33L308.19 39.14a32.16 32.16 0 0 0-40.38 0L35 229.16a8 8 0 0 0-1.16 11.24l10.1 12.41a8 8 0 0 0 11.2 1.19L96 220.62v243a16 16 0 0 0 16 16h128a16 16 0 0 0 16-16v-128l64 .3V464a16 16 0 0 0 16 16l128-.33a16 16 0 0 0 16-16V220.62L520.86 254a8 8 0 0 0 11.25-1.16l10.1-12.41a8 8 0 0 0-1.21-11.27zm-93.11 218.59h.1l-96 .3V319.88a16.05 16.05 0 0 0-15.95-16l-96-.27a16 16 0 0 0-16.05 16v128.14H128V194.51L288 63.94l160 130.57z"></path>
</svg>```

bakerac4:
@nrshapiro just confirming before I start, you want me to change the background color of the tooltips?

nrshapiro:
[...]

PR #513, opened 2021-05-17
Fixing tooltips

### LINKED ISSUE #512: Tooltips clipped?
This isn't happening on my system, but I'll report what happened in the PEP board meeting when they were demonstrating the new, longer tooltips.

In particular, the one that's uppermost: the SmartSearch tip, is going off the screen at the top and clipped.

Here's the screen capture they sent.  I am only testing now on the PC--my Mac laptop is on loan.  But I suspect they all use Macs.

2021-05-17_09h39_08

bakerac4:
Fixes #512 and #503

### DIFF
--- a/pep/app/styles/components/_ember-tooltip.scss
+++ b/pep/app/styles/components/_ember-tooltip.scss
@@ -1,4 +1,31 @@
 .ember-popover {
+    background-color: #2f3136;
+    border-color: darken(#2f3136, 5%);
+    color: #dbdddc;
+    &[x-placement^='right'] {
+        .ember-popover-arrow {
+            border-right-color: darken(#2f3136, 5%) !important;
+        }
+    }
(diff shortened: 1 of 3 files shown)


pep/app/styles/components/_ember-tooltip.scss
```diff
@@ -1,4 +1,31 @@
...
+    }
+    &[x-placement^='bottom'] {
+        .ember-popover-arrow {
+            border-bottom-color: darken(#2f3136, 5%) !important;
+        }
+    }
+    a {
+        color: #6a81ae;
+    }
+
```
billdami, line 28: do all these colors need theme specific vars, or are they the same for all?

bakerac4: They are the same for all

PR #550, opened 2021-07-22
Story/tp68442 redirect

bakerac4:


pep/node/server.js
```diff
@@ -13,6 +13,21 @@ module.exports = function(emberDistPath) {
...
+            res.redirect(301, `/browse/${journal}/volumes/${volume}`);
+        } else {
+            res.redirect(301, `/browse/${journal}/volumes`);
+        }
+    });
+
+    server.all('/document.php', (req, res) => {
+        const id = req.query.id;
+        res.redirect(301, `/browse/document/${id}`);
+    });
```
billdami, line 29: won't hold this up as I'm sure these changes are a priority, but @adistasio @lhorne-gavant I wonder if redirects like this would be better off in the api gateway/lambda layer, if we can config any sort of mod_rewrite-like rules easily there? 

Just so these old requests dont need to get to and spin up the fastboot lambda, only to be redirected to another fastboot lambda request.

billdami (approved):

PR #594, opened 2022-01-05
Revised printing rules

### LINKED ISSUE #591: Revised Print/Download Restriction Feature
Continuing from #561 

We would like to leave the print/download icons in place, but disable them.  If the user hovers or clicks on the "grayed" icons, we would like a message explaining that printing/download is not available for the current document due to copyright security (configurable with the other hover messages).

Since I am planning to push to production the server and client on January 22, and we need testing time on Stage, we'd like this done by Jan 7.

@davidtuckett

bakerac4:
Fixes #591

### DIFF
--- a/pep/app/constants/configuration.ts
+++ b/pep/app/constants/configuration.ts
@@ -138,6 +138,14 @@ export interface ContentConfiguration {
         };
         adminSpecifiedInformationItems: AdminSpecifiedInformation[];
         notificationMessage?: string;
+        access: {
+            print: {
+                help: string;
+            };
+            download: {
+                help: string;
+            };
+        };
     };
     home: {
         intro: {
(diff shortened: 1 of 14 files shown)


nrshapiro:
@bakerac4 @rickardnyman 

In a TOC the icons look like this:
image

when reading a document that can be printed or download (my admin account) they look like this:

image

When reading a book section, that apparently cannot be printed, they look like: 

image

Note the printer icon is hollow in both cases.  Following from the TOC, shouldn't it be solid when I can print, and hollow when I can't?

The hover/click message works well when not allowed.

Note: I'm just commenting on the icons.  I still need to look at whether they are disabled in the right cases based on the server data.

billdami (approved):

PR #615, opened 2022-01-24
Fixing auth model refresh issue

### LINKED ISSUE #547: Reading an abstract (not signed in), after signing in, article still shows available to subscribers
1) Login to an account where the preference is set to:
    image

2) Logout.  Note that the preference will still be set while logged out.

3) Search for an article, e.g. in this case just type GW.010.0248A into smartsearch

4) You'll see:
image

5) Now Log in to an account with PEPArchive access.

6) You'll see:
image
7) You can do a Refresh (F5) to show the article, but the interface should show the full article as soon as you log in.  Most users would be stuck at this point wondering what went wrong.

NOTES:
a) This doesn't happen to users if the preference isn't set as above since the won't be in read mode anyway.
b) This would apply to any user who is timed out, or who logs out and searches while logged out, if their preference is set that way.

bakerac4:
Fixes #547

### DIFF
--- a/pep/app/pods/browse/read/controller.ts
+++ b/pep/app/pods/browse/read/controller.ts
@@ -94,15 +94,14 @@ export default class BrowseRead extends Controller {
      *
      * @memberof SearchRead
      */
-    setup() {
+    addAuthenticatedListener() {
         this.session.on('authenticationAndSetupSucceeded', this.onAuthenticated);
     }
 
     /**
      * Removes the auth succeeded event listener on destroy of controller
      */
-    willDestroy() {
-        super.willDestroy();
+    removeAuthenticateListener() {
         this.session.off('authenticationAndSetupSucceeded', this.onAuthenticated);
     }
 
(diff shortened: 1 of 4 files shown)


pep/app/pods/browse/read/controller.ts
```diff
@@ -89,6 +89,23 @@ export default class BrowseRead extends Controller {
...
+        this.session.on('authenticationAndSetupSucceeded', this.onAuthenticated);
+    }
+
+    /**
+     * Removes the auth succeeded event listener on destroy of controller
+     */
+    willDestroy() {
+        super.willDestroy();
+        this.session.off('authenticationAndSetupSucceeded', this.onAuthenticated);
+    }
```
billdami, line 107: since controllers are singletons they shouldn't normally be destroyed over the life of the app right? If thats the case, I wonder if you need to remove the listener on like the `didTransition` of the route or something too, otherwise you might be setting up multiple duplicate listeners each time the route is visited.

bakerac4: Yes good point, honestly should use 'resetController' on the route

billdami (approved):

PR #653, opened 2022-05-27
Story/gh595 embargo

### LINKED ISSUE #595: TOC embargo year information changed January 1, but the "wall" moves January 22 (and can vary)
The automatic TOC embargo year information on the TOC for each journal has changed.  I believe it's set to use the embargo data in Admin and compute the available years.  So now, in the TOC for JCP, which has a three year wall:, it shows:

image

I believe this changed January 1.  But the "wall" typically moves January 22 (and can vary, if we fall behind in reprocessing the data.)

So what we probably need to do is to add a embargo release date to "Admin", which can be set to Jan 22, but changed each year as needed.  

The slowdown in moving the wall is mainly the new data, so we could also consider moving the wall on Jan 1, and releasing the new data on January 22.  That's not too complicated to explain. :)

bakerac4:
Fixes #595


pep/app/pods/admin/general/template.hbs
```diff
@@ -363,6 +359,28 @@
                         </div>
                     </div>
                 </div>
+                <div class="card mb-4">
+                    <div class="card-body">
+                        <div class="form-row">
+                            <div class="col-12">
+                                <h4 class="d-flex justify-content-between align-items-center border-bottom mb-3 pb-1">
+                                    Embargo Date
```
billdami, line 367: should this be translated?

billdami (approved):

PR #679, opened 2022-11-30
UI quick-fixes

jordanallen-dev:
Series of minor UI changes in response to the UX-Thinkers meeting

- Add missing tooltips / titles
- Link logo to homepage instead of support page
- Add support link to navbar and mobile sidebar
- Remove unused development mode
- Open browse drawer when browse link is clicked
- Remove unused "last read" functionality


nrshapiro:
If I understand this correctly, the development mode is what I believe I need to run the client on Windows here.  Though of late, I've had problems installing it here anyway.  But I can work on server issues much better if I can install the client on my Windows 11 PC.

jordanallen-dev:
> If I understand this correctly, the development mode is what I believe I need to run the client on Windows here. Though of late, I've had problems installing it here anyway. But I can work on server issues much better if I can install the client on my Windows 11 PC.

You can't run the client on your Windows machine?

That's definitely problematic. I'm not sure how this dev mode would've helped

I can try and help you work it out on a call if you'd like?

PR #681, opened 2022-12-01
feature/disqus-sso

jordanallen-dev:
Implement Disqus SSO service

Automatically create and authenticate a Disqus account for a signed in PEP user

Prevent sign out of Disqus without signing out of PEP


jordanallen-dev:
Blocked waiting for Zedra to release Disqus SSO signature endpoint to production

PR #693, opened 2022-12-12
feature/render-archived-document-versions

jordanallen-dev:
Render version history when present

Access archived documents via the new archival endpoint


jordanallen-dev:
Blocked waiting for fix to version history on latest version:

image

